PR #3, opened 2017-01-19
Initial Review

nvanbenschoten:
Various comments, but only reviewed part of the code so far.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/3)

### DIFF
--- a/context.go
+++ b/context.go
@@ -22,6 +22,10 @@ import (
 )
 
 // Context maintains options for Decimal operations.
+//
+// REVIEW: can this be used concurrently? If not, we may be able to cache a few
+// temporary allocations used in different routines (e.g. dividend and divisor).
+// If so, say so.
 type Context struct {
 	// Precision is the number of places to round during rounding.
 	Precision uint32
@@ -44,6 +48,8 @@ type Context struct {
 const (
 	// DefaultTraps is the default trap set used by BaseContext. It traps all
 	// flags except Inexact and Rounded.
+	//
+	// REVIEW: nit, why two on one line here?
 	DefaultTraps = SystemOverflow | SystemUnderflow |
 		Overflow |
 		Underflow |
@@ -53,10 +59,13 @@ const (
 		DivisionImpossible |
 		InvalidOperation
 
+	// REVIEW: same error string naming questions
 	errZeroPrecision = "Context may not have 0 Precision for this operation"
 )
 
 // BaseContext is a useful default Context.
+//
+// REVIEW: comment that it should not be mutated.
 var BaseContext = Context{
 	// Disable rounding.
 	Precision: 0,
@@ -74,11 +83,15 @@ func (c *Context) WithPrecision(p uint32) *Context {
 	return &r
 }
 
+// REVIEW: this deserves a comment.
 func (c *Context) goError(flags Condition) (Condition, error) {
 	return flags.GoError(c.Traps)
 }
 
 // Add sets d to the sum x+y.
+//
+// REVIEW: do these work if d == x, d == y, x == y, or d == x == y?
+// Same question for all. Should be documented.
 func (c *Context) Add(d, x, y *Decimal) (Condition, error) {
 	a, b, s, err := upscale(x, y)
 	if err != nil {
@@ -122,7 +135,10 @@ func (c *Context) Mul(d, x, y *Decimal) (Condition, error) {
 }
 
 // Quo sets d to the quotient x/y for y != 0. c's Precision must be > 0.
+// REVIEW: s/c's Precision/c.Precision/
 func (c *Context) Quo(d, x, y *Decimal) (Condition, error) {
+	// REVIEW: should we return an error for Precision < 0
+	// REVIEW: Seems like a good case for a switch statement (disjoint cases)
 	if c.Precision == 0 {
 		// 0 precision is disallowed because we compute the required number of digits
 		// during the 10**x calculation using the precision.
@@ -145,6 +161,7 @@ func (c *Context) Quo(d, x, y *Decimal) (Condition, error) {
 	// An integer variable, adjust, is initialized to 0.
 	var adjust int64
 	// The result coefficient is initialized to 0.
+	// REVIEW: could we use d here to avoid an allocation if (d != nil) && (d != x) && (d != y)?
 	quo := new(Decimal)
 	var res Condition
 	var diff int64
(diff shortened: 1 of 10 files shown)


mjibson:
Oh! I wasn't subscribed to this repo and didn't see this until just now. Thanks for the initial comments, will look at them soon.

context.go
```diff
@@ -122,7 +135,10 @@ func (c *Context) Mul(d, x, y *Decimal) (Condition, error) {
 }
 
 // Quo sets d to the quotient x/y for y != 0. c's Precision must be > 0.
+// REVIEW: s/c's Precision/c.Precision/
 func (c *Context) Quo(d, x, y *Decimal) (Condition, error) {
+	// REVIEW: should we return an error for Precision < 0
```
nvanbenschoten: Right, but we're not returning an error, are we?

decimal.go
```diff
@@ -35,13 +35,17 @@ type Decimal struct {
 }
 
 // New creates a new decimal with the given coefficient and exponent.
+//
+// REVIEW: I wonder if a Make function would also be useful. Ideally you wouldn't
+// always need to allocate the outer struct onto the heap.
```
nvanbenschoten: No I mean:
```
func Make(coeff int64, exponent int32) Decimal {
	return Decimal{
		Coeff:    *big.NewInt(coeff),
		Exponent: exponent,
	}
}
```
There might be cases where forcing the allocation is unnecessary.

decimal.go
```diff
@@ -65,7 +71,10 @@ func newFromString(s string) (coeff *big.Int, exps []int64, err error) {
 
 // NewFromString creates a new decimal from s. It has no restrictions on
 // exponents or precision.
+//
+// REVIEW: Do we ever test error conditions here? I can imagine a number.
 func NewFromString(s string) (*Decimal, error) {
+	// REVIEW: why is this different from BaseContext.NewFromString(s)?
```
nvanbenschoten: Yep, the implementation was all I was referring to. I agree on the API standpoint.

decimal.go
```diff
@@ -107,10 +117,18 @@ func (c *Context) NewFromString(s string) (*Decimal, Condition, error) {
 }
 
 // String is a wrapper of ToSci.
+// REVIEW: why?
```
nvanbenschoten: I just meant why have `ToSci` in the first place then. I didn't see `ToStandard` before though so feel free to ignore.

decimal.go
```diff
@@ -107,10 +117,18 @@ func (c *Context) NewFromString(s string) (*Decimal, Condition, error) {
...
+// REVIEW: Please please please support an Append(buf []byte) []byte method to avoid
+//   allocations! That was one of my biggest issues with other libraries when dealing with
+//   encoding/decoding code. In general, a major design consideration here should be making
+//   sure we have no (or minimal) allocation fast-paths for encoding/decoding. See util/encoding/decimal.go.
```
nvanbenschoten: Awesome! Glad to see that speedup.

nvanbenschoten:
Added a second review. Only need to finish reviewing `decimal.go` and `context.go` at this point.

nvanbenschoten:
Added a third review. I have now made it through all files.

context.go
mjibson: Oh precision < 0, I misread that is == 0. Precision is a uint32, so it can't be < 0.

decimal.go
mjibson: ToSci in the first place because the GDA spec has both a ToSci and ToEng formatter (http://speleotrove.com/decimal/daconvs.html), and I wanted our API to reflect closely that spec. I haven't done ToEng yet, which is why ToSci alone seems weird.

decimal.go
mjibson: Ah. I'd like to add this in the future when we find a place where it is clearly beneficial.

error_test.go
```diff
@@ -32,6 +33,7 @@ func TestErrDecimal(t *testing.T) {
 	ed.Round(a, a)
 }
 
+// REVIEW: this is testing something more specific than "NewErrDecimal", right?
```
mjibson: it's really a test of c.WithPrecision generating a copy. Added a comment and renamed.

round.go
```diff
@@ -18,6 +18,7 @@ import "math/big"
 
 // Round sets d to rounded x. If d has zero Precision, no rounding will
 // occur. If d has no Rounding specified, RoundHalfUp is used.
+// REVIEW: does d have a Precision or a Rounding, or does c?
```
mjibson: c. fixed.

round.go
```diff
@@ -66,6 +67,7 @@ func (r Rounder) Round(c *Context, d, x *Decimal) Condition {
 		if m.Sign() != 0 {
 			res |= Inexact
 			m.Abs(m)
+			// REVIEW: use NewWithBigInt (which we should have).
```
mjibson: done

round.go
```diff
@@ -79,6 +81,7 @@ func (r Rounder) Round(c *Context, d, x *Decimal) Condition {
 
 // roundAddOne adds 1 to abs(b).
 func roundAddOne(b *big.Int, diff *int64) {
+	// REVIEW: any way we can avoid calling this twice?
```
mjibson: I'm not aware of a better way to determine if there was a 9 -> 10 rollover here which added another decimal digit. At least it uses a lookup table for up to 128 bits so it's generally constant time..

round.go
```diff
@@ -92,6 +95,7 @@ func roundAddOne(b *big.Int, diff *int64) {
 	}
 }
 
+// REVIEW: all of these should be tested.
```
mjibson: They are: https://github.com/cockroachdb/apd/blob/master/testdata/rounding.decTest

error.go
```diff
@@ -24,6 +24,9 @@ func NewErrDecimal(c *Context) *ErrDecimal {
 // ErrDecimal performs operations on decimals and collects errors during
 // operations. If an error is already set, the operation is skipped. Designed to
 // be used for many operations in a row, with a single error check at the end.
+// REVIEW: I dont love this name as it's kind of misleading. ErrDecimal sounds like
```
mjibson: I agree but have no idea. ErrAccumulator? Dunno.

mjibson:
I think I've incorporated all this feedback now. Can you close this PR if you agree?

nvanbenschoten:
Yep, thanks for addressing these comments.

PR #4, opened 2017-01-22
address review comments

mjibson:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/4)

### DIFF
--- a/context.go
+++ b/context.go
@@ -21,7 +21,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Context maintains options for Decimal operations.
+// Context maintains options for Decimal operations. It can safely be used
+// concurrently, but not modified concurrently.
 type Context struct {
 	// Precision is the number of places to round during rounding.
 	Precision uint32
@@ -43,7 +44,8 @@ type Context struct {
 
 const (
 	// DefaultTraps is the default trap set used by BaseContext.
-	DefaultTraps = SystemOverflow | SystemUnderflow |
+	DefaultTraps = SystemOverflow |
+		SystemUnderflow |
 		Overflow |
 		Underflow |
 		Subnormal |
@@ -52,10 +54,10 @@ const (
 		DivisionImpossible |
 		InvalidOperation
 
-	errZeroPrecision = "Context may not have 0 Precision for this operation"
+	errZeroPrecisionStr = "Context may not have 0 Precision for this operation"
 )
 
-// BaseContext is a useful default Context.
+// BaseContext is a useful default Context. Should not be mutated.
 var BaseContext = Context{
 	// Disable rounding.
 	Precision: 0,
@@ -73,6 +75,7 @@ func (c *Context) WithPrecision(p uint32) *Context {
 	return &r
 }
 
+// goError converts flags into an error based on c.Traps.
 func (c *Context) goError(flags Condition) (Condition, error) {
 	return flags.GoError(c.Traps)
 }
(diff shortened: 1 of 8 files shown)


context.go
```diff
@@ -323,7 +328,7 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
 	tmp := new(Decimal)
 	// The algorithm in the paper says to use c.Precision + 2. 7 instead of 2
```
arjunravinarayan: Can you add a citation to the algorithm/paper being followed in the function header comment? Otherwise this comment is too cryptic to follow. You should probably do this for all papers followed

round.go
```diff
@@ -16,10 +16,10 @@ package apd
 
 import "math/big"
 
-// Round sets d to rounded x. If d has zero Precision, no rounding will
-// occur. If d has no Rounding specified, RoundHalfUp is used.
+// Round sets d to rounded x. If c has zero Precision, no rounding will
```
arjunravinarayan: "Round sets d to rounded x, rounded to the precision specified by c." would be better.

context.go
```diff
@@ -117,15 +120,17 @@ func (c *Context) Mul(d, x, y *Decimal) (Condition, error) {
...
+// it was exact by checking the Inexact flag on the return Condition.
 func (c *Context) Quo(d, x, y *Decimal) (Condition, error) {
 	if c.Precision == 0 {
 		// 0 precision is disallowed because we compute the required number of digits
 		// during the 10**x calculation using the precision.
-		return 0, errors.New(errZeroPrecision)
+		return 0, errors.New(errZeroPrecisionStr)
 	}
 	if c.Precision > 5000 {
 		return 0, errors.New("Quo requires Precision <= 5000")
```
arjunravinarayan: When you add artificial limits like this, add a comment saying why. The answer is almost always "performance" but its good to have that written down somewhere.

context.go
```diff
@@ -339,9 +344,8 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
 		// approx = 0.5 * (approx + f / approx)
 		ed.Mul(approx, tmp, decimalHalf)
 	}
-	p = c.Precision
-	nc.Precision = p
-	dp := int32(p)
+	nc.Precision = c.Precision
+	dp := int32(c.Precision)
 	approxsubhalf := new(Decimal)
```
arjunravinarayan: cryptic variable name needs a comment saying what it is

context.go
```diff
@@ -389,7 +393,7 @@ func (c *Context) Cbrt(d, x *Decimal) (Condition, error) {
 
 	z := new(Decimal).Set(x)
 	nc := BaseContext.WithPrecision(c.Precision*2 + 2)
-	ed := NewErrDecimal(nc)
+	ed := MakeErrDecimal(nc)
 	exp8 := 0
 
 	// Follow Ken Turkowski paper:
```
arjunravinarayan: Full citation?

context.go
```diff
@@ -463,7 +467,7 @@ func (c *Context) Ln(d, x *Decimal) (Condition, error) {
```
arjunravinarayan: E.g. the comment at the start of this function is a wonderful example.

context.go
mjibson: done

context.go
mjibson: This has already been done, but isn't represented in this branch.

context.go
mjibson: It's the same name as used in the paper. I think it's best to keep it as is because of that.

context.go
mjibson: added

round.go
mjibson: done

PR #5, opened 2017-01-27
Full tests

mjibson:
This PR adds the remainder of the full decTest tests which come with many small fixes. The first 3 commits are the same as the other PR.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/5)


context.go
```diff
@@ -80,6 +80,10 @@ func (c *Context) goError(flags Condition) (Condition, error) {
 	return flags.GoError(c.Traps)
 }
 
+func (c *Context) Etiny() int32 {
```
arjunravinarayan: This function name is cryptic; it needs a comment as to its purpose.

mjibson: Commented. I also unexported this because it'll be rare that users need it.

PR #7, opened 2017-02-03
ln: minor improvements

RaduBerinde:
Hi Matt,

The Ln code looks good to me! I just have this suggested minor improvement to replace `Mul` by `decimalTwo` with an `Add` (should be a bit faster). I couldn't run the benchmark to verify if we see an improvement though.

I think we could stop the loop when the delta (tmp2) becomes small enough (less than `10^-c.Precision` or so). We would also optimize out the `Sub` in `loop.done()` since we already know the delta. I will look at that.

By the way, I don't quite understand the logic in `loop.done()`; we "stall" when `l.delta.Cmp(l.prevDelta) == 0`? Why? I would expect in most iterative algorithms the delta to keep decreasing.. So we should stop when it becomes 0 or smaller than a certain value no?

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/7)

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -1,0 +1,87 @@
+// Copyright 2017 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package apd
+
+import (
+	"bytes"
+	"fmt"
+	"math/rand"
+	"testing"
+)
+
+// runBenches benchmarks a given function on random decimals on combinations of
+// three parameters:
+//
+//   precision:    desired output precision
+//   inScale:      the scale of the input decimal: the absolute value will be between
+//                 10^inScale and 10^(inScale+1)
+//   inNumDigits:  number of digits in the input decimal; if negative the number
+//                 will be negative and the number of digits are the absolute value.
+func runBenches(
+	b *testing.B, precision, inScale, inNumDigits []int, fn func(*testing.B, *Context, *Decimal),
+) {
+	for _, p := range precision {
+		for _, s := range inScale {
+			for _, d := range inNumDigits {
+				numDigits := d
+				negative := false
+				if d < 0 {
+					numDigits = -d
+					negative = true
+				}
+
+				// Generate some random numbers with the given number of digits.
+				nums := make([]Decimal, 20)
+				for i := range nums {
+					var buf bytes.Buffer
+					if negative {
+						buf.WriteByte('-')
+					}
+					buf.WriteByte('1' + byte(rand.Intn(9)))
+					for j := 1; j < numDigits; j++ {
+						buf.WriteByte('0' + byte(rand.Intn(10)))
+					}
+					if _, err := nums[i].SetString(buf.String()); err != nil {
+						b.Fatal(err)
+					}
+					nums[i].SetExponent(int32(s - d))
+				}
+				b.Run(
+					fmt.Sprintf("P%d/S%d/D%d", p, s, d),
+					func(b *testing.B) {
+						ctx := BaseContext.WithPrecision(uint32(p))
+						for i := 0; i <= b.N; i++ {
+							fn(b, ctx, &nums[i%len(nums)])
+						}
+					},
+				)
+			}
+		}
+	}
+}
+
+func BenchmarkLn(b *testing.B) {
+	precision := []int{2, 10, 100}
+	scale := []int{-100, -10, -2, 2, 10, 100}
+	digits := []int{2, 10, 100}
+	runBenches(
+		b, precision, scale, digits,
+		func(b *testing.B, ctx *Context, x *Decimal) {
+			if _, err := ctx.Ln(&Decimal{}, x); err != nil {
+				b.Fatal(err)
+			}
+		},
+	)
+}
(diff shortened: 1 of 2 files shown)


RaduBerinde:
I will also think about the necessary precision and what would be the most accurate way to calculate it. The halving/doubling steps make this harder to calculate.

One other thing - the halving/doubling steps might take a long time for large or very small numbers. It may be beneficial to add some steps that multiply by a (precalculated) 2^10 or 2^-10 or so to cut down on these steps. We need benchmarks across a wide range of precisions and magnitudes if we are to tune these properly though.

RaduBerinde:
Or even better, we could "shift" the number by 10^x to make it be between 0 and 1 first (just a manipulation of exponents)..

RaduBerinde:
Does `ed.Mul(tmp2, m, decimalLog2)` multiply using all 3000 digits of `decimalLog2` before rounding down? If so that could be by far the slowest part of the function whenever we only need a handful of digits of precision..  What might help here would be a facility that precalculates rounded-down versions of these long constants (e.g. for precision, 1, 2, 4, 8, 16, ...) so you can just look up a version of `decimalLog2` that's not much longer than what you need.

RaduBerinde:
Added some benchmarks. At first glance, the function seems pretty slow even for easy cases (.4ms for precision=2 and input with 2 digits). We will probably first need to characterize `Exp` though before working on improving `Ln`.

mjibson:
I'm not sure why the loop code compares delta like that. We copied the code almost verbatim from https://github.com/robpike/ivy/blob/master/value/loop.go. If we can find a faster way I'm all for it. I think your analysis about tmp2 may be correct. Give it a try and see if the benchmarks improve?

I think shifting by 10 using exponent manipulation should work fine, yes. That sounds like a big win.

Yes, decimalLog2 multiplies by all 3000 digits. We do need some better way to get the digits we need? We could also consider using substrings of the original log2 string and making a decimal on demand? Although that sounds like it could be slower.

bench_test.go
```diff
@@ -0,0 +1,87 @@
...
+					}
+					if _, err := nums[i].SetString(buf.String()); err != nil {
+						b.Fatal(err)
+					}
+					nums[i].SetExponent(int32(s - d))
+				}
+				b.Run(
+					fmt.Sprintf("P%d/S%d/D%d", p, s, d),
+					func(b *testing.B) {
+						ctx := BaseContext.WithPrecision(uint32(p))
```
mjibson: This should be created only once per precision.

RaduBerinde:
> I'm not sure why the loop code compares delta like that. We copied the code almost verbatim from https://github.com/robpike/ivy/blob/master/value/loop.go. If we can find a faster way I'm all for it.

Seems incorrect to me.. If my first three steps of the iteration all increase the value by 10, why should we stop? It seems like they thought we were computing the delta to some fixed value but that's not the case. I'll open an issue in that repo.


> Yes, decimalLog2 multiplies by all 3000 digits. We do need some better way to get the digits we need? We could also consider using substrings of the original log2 string and making a decimal on demand? Although that sounds like it could be slower.

We could precalculate Decimals for log(3000) precisions and store them in a look up table; that way we don't calculate anything on demand but we never use something with more than 2x digits than we need.

mjibson:
Yes, I like that log precompute idea.

RaduBerinde:
Thanks. I added a benchmark for Exp as well.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[bench_test.go, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/7#-Kc3rrYAq6MP7eYLH5bH-r2-64:-Kc449_vKvLy35WYGvOc:b-icfism) ([raw file](https://github.com/cockroachdb/apd/blob/9f70fb262800460e3537a8028c757ce4734a8e45/bench_test.go#L64)):*

This should be created only once per precision.
</blockquote></details>

Are you suggesting moving the outer `for` under the others? I don't like it because it generates the results in a different order, and I want to keep `Pxx` in front (it seems to be the more important factor for performance).

I can generate a table of numbers and reuse them, but what's the point? This is nothing compared to the actual benchmark..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/7#-:-Kc44l_eAF1_sxHD3Vrs:b7klfd1)*

mjibson:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[bench_test.go, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/7#-Kc3rrYAq6MP7eYLH5bH-r2-64:-Kc6QV7Saa3nI6M76uBF:bogl0tt) ([raw file](https://github.com/cockroachdb/apd/blob/9f70fb262800460e3537a8028c757ce4734a8e45/bench_test.go#L64)):*

Are you suggesting moving the outer `for` under the others? I don't like it because it generates the results in a different order, and I want to keep `Pxx` in front (it seems to be the more important factor for performance).

I can generate a table of numbers and reuse them, but what's the point? This is nothing compared to the actual benchmark..
</blockquote></details>

ctx should be created right after the `for _, p := range precision` line because ctx doesn't change, so it can be reused as long as the precision stays the same. This is just a small optimization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/7)*

RaduBerinde:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[bench_test.go, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/7#-Kc3rrYAq6MP7eYLH5bH-r2-64:-KcED3RopcsUhnZh9vr5:b-1yx5a2) ([raw file](https://github.com/cockroachdb/apd/blob/9f70fb262800460e3537a8028c757ce4734a8e45/bench_test.go#L64)):*

ctx should be created right after the `for _, p := range precision` line because ctx doesn't change, so it can be reused as long as the precision stays the same. This is just a small optimization.
</blockquote></details>

Ah, thanks, for some reason I thought the comment was referring to `nums[..]` but I misread. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/7)*

PR #11, opened 2017-02-06
Test cbrt

mjibson:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/11)

### DIFF
--- a/const.go
+++ b/const.go
@@ -62,7 +62,9 @@ type constWithPrecision struct {
 
 func makeConstWithPrecision(strVal string) *constWithPrecision {
 	c := &constWithPrecision{}
-	c.unrounded.SetString(strVal)
+	if _, err := c.unrounded.SetString(strVal); err != nil {
+		panic(err)
+	}
 	// The length of the string might be one higher than the availble precision
 	// (because of the decimal point), but that's ok.
 	maxPrec := uint32(len(strVal))
--- a/context.go
+++ b/context.go
@@ -454,12 +454,10 @@ func (c *Context) Cbrt(d, x *Decimal) (Condition, error) {
 		ed.Mul(z, z, decimalTwo)
 	}
 
-	z0.Set(z)
-
 	// Loop until convergence.
-	for loop := nc.newLoop("cbrt", z, c.Precision, 1); false; {
+	for loop := nc.newLoop("cbrt", z, c.Precision+1, 1); ; {
 		// z = (2.0 * z0 +  x / (z0 * z0) ) / 3.0;
-		z.Set(z0)
+		z0.Set(z)
 		ed.Mul(z, z, z0)
 		ed.Quo(z, x, z)
 		ed.Add(z, z, z0)
@@ -474,13 +472,23 @@ func (c *Context) Cbrt(d, x *Decimal) (Condition, error) {
 		} else if done {
 			break
 		}
-		z0.Set(z)
 	}
+
+	res, err := c.Round(d, z)
+
+	// Set z = d^3 to check for exactness.
+	ed.Mul(z, d, d)
+	ed.Mul(z, z, d)
 
 	if err := ed.Err(); err != nil {
 		return 0, err
 	}
-	return c.Round(d, z)
+
+	// Result is exact
+	if x.Cmp(z) == 0 {
+		return 0, nil
+	}
+	return res, err
 }
 
 // Ln sets d to the natural log of x.
--- a/gda_test.go
+++ b/gda_test.go
@@ -225,6 +225,9 @@ var GDAfiles = []string{
 	"subtract",
 	"tointegral",
 	"tointegralx",
+
+	// non-GDA tests
+	"cuberoot-apd",
 }
 
 func TestGDA(t *testing.T) {
--- a/testdata/cuberoot-apd.decTest
+++ b/testdata/cuberoot-apd.decTest
@@ -89,3 +89,8 @@ cbtx086 cuberoot 343 -> 7
 cbtx087 cuberoot 512 -> 8
 cbtx088 cuberoot 729 -> 9
 cbtx089 cuberoot 1000 -> 10
+
+-- others
+precision: 5
+
+cbtx100 cuberoot 1.0001 -> 1.0000 Inexact Rounded


RaduBerinde:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[context.go, line 460 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGr6kGUGMhKVlPfMNm:-KcGr6kGUGMhKVlPfMNn:bfxgmwo) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L460)):*
> ```Go
> 
> 	// Loop until convergence.
> 	for loop := nc.newLoop("cbrt", z, c.Precision, 1); ; {
> ```

We may need `c.Precision+1` here (not sure if this is related to the new check below).

And sorry about the false, I was experimenting and I guess I accidentally left it there.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcGsPcgf_Do4JX5M8ek:b-i4vuq1) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*
> ```Go
> 	}
> 
> 	// Cbrt has a flawed inexact determination. Due to how the loop termination
> ```

Could we check if we have an "exact" flag, and in that case multiply z by itself two times and see if it exactly equals the input?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

mjibson:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[context.go, line 460 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGr6kGUGMhKVlPfMNm:-KcHSqk1-d4_A2CQi7d2:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L460)):*

We may need `c.Precision+1` here (not sure if this is related to the new check below).

And sorry about the false, I was experimenting and I guess I accidentally left it there.
</blockquote></details>

Done.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcHSrTjlw1F20B5HSU6:bc9kyaf) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

Could we check if we have an "exact" flag, and in that case multiply z by itself two times and see if it exactly equals the input?
</blockquote></details>

Inexact occurs during rounding when some non-zero digits are removed. Since this problem is caused by non-zero digits, we can't check for an exact flag because it's known to be inexact. For example `cbrt(1)` was getting something like `1.000000000000013`. After rounding this would set the Inexact flag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

RaduBerinde:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcIPXdSb2TL-yb6koI_:bh7m911) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

Inexact occurs during rounding when some non-zero digits are removed. Since this problem is caused by non-zero digits, we can't check for an exact flag because it's known to be inexact. For example `cbrt(1)` was getting something like `1.000000000000013`. After rounding this would set the Inexact flag.
</blockquote></details>

Ah, but then just because we were able to trim some zeroes after rounding, why do we decide that the result is exact? It could end in zeroes even if it isn't..

E.g. input 1.0001 with precision 5 should result in `1.0000` after rounding, which is correct for the precision (exact answer is ~ 1.00003333222), but not exact.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

mjibson:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcJTz7ADd3k0zG2Nm79:b-mz2cps) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

Ah, but then just because we were able to trim some zeroes after rounding, why do we decide that the result is exact? It could end in zeroes even if it isn't..

E.g. input 1.0001 with precision 5 should result in `1.0000` after rounding, which is correct for the precision (exact answer is ~ 1.00003333222), but not exact.
</blockquote></details>

Yes, the situation you describe is exactly the problem this comment is describing. It is possible that this method needs the old loop code. Or maybe we need more tests like the one you list here and to figure out an algorithm that always works.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

RaduBerinde:
*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcJWtvMZXgR_0MPtP71:b3bz5h8) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

Yes, the situation you describe is exactly the problem this comment is describing. It is possible that this method needs the old loop code. Or maybe we need more tests like the one you list here and to figure out an algorithm that always works.
</blockquote></details>

I think with any loop termination code the problem above could happen.. So after we do the round and reduce, can't we multiply `d` with itself two times to see if it exactly matches x?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

RaduBerinde:
*[context.go, line 477 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcJZTMtucXpFd28JGx0:-KcJZTMtucXpFd28JGx1:b-we28r3) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L477)):*
> ```Go
> 			break
> 		}
> 		z0.Set(z)
> ```

These `z0.Set(z)` and `z.Set(z0)` are confusing. Can we replace all three by just a `z0.Set(z)` at the beginning of the loop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

mjibson:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[context.go, line 477 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcJZTMtucXpFd28JGx0:-KcJi7GVYCr6KesFaDVh:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L477)):*

These `z0.Set(z)` and `z.Set(z0)` are confusing. Can we replace all three by just a `z0.Set(z)` at the beginning of the loop?
</blockquote></details>

Done.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcJhRUR0_HLTveOp0Z_:b49smxf) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

I think with any loop termination code the problem above could happen.. So after we do the round and reduce, can't we multiply `d` with itself two times to see if it exactly matches x?
</blockquote></details>

Yes, this works. Much better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11)*

RaduBerinde:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 484 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/11#-KcGsPcgf_Do4JX5M8ej:-KcK-lxcUn83xwsVapk_:b-83sjoh) ([raw file](https://github.com/cockroachdb/apd/blob/5185518c92c17ae62d8b6b4a50e1a120973c0baa/context.go#L484)):*

Yes, this works. Much better.
</blockquote></details>

Nice!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/11#-:-KcK-mfCYkS51r8nxl_H:bnfp4nl)*

PR #18, opened 2017-02-13
Update travis to use point releases

mjibson:
Before it was stuck to go1.8rc1 and 1.7.4.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/18)


mjibson:
Nope. Travis is just slow to update I guess.

PR #20, opened 2017-02-16
use uint64 in reduce when possible

mjibson:
```
name          old time/op  new time/op  delta
GDA/reduce-4   215µs ±30%   124µs ±10%  -42.48%  (p=0.008 n=5+5)
```

Idea from: https://github.com/cockroachdb/cockroach/blob/0ba52ff84a3e67efe3532b9119058aab8231478a/pkg/util/encoding/decimal.go#L773

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/20)

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -462,12 +462,38 @@ func (d *Decimal) Abs(x *Decimal) *Decimal {
 
 // Reduce sets d to x with all trailing zeros removed and returns d.
 func (d *Decimal) Reduce(x *Decimal) *Decimal {
-	if x.Sign() == 0 {
+	neg := false
+	switch x.Sign() {
+	case 0:
 		d.SetCoefficient(0)
 		d.Exponent = 0
 		return d
+	case -1:
+		neg = true
 	}
 	d.Set(x)
+
+	// Use a uint64 for the division if possible.
+	if d.Coeff.BitLen() <= 64 {
+		i := d.Coeff.Uint64()
+		e := d.Exponent
+		for i >= 10000 && i%10000 == 0 {
+			i /= 10000
+			e += 4
+		}
+		for i%10 == 0 {
+			i /= 10
+			e++
+		}
+		if e != d.Exponent {
+			d.Exponent = e
+			d.Coeff.SetUint64(i)
+			if neg {
+				d.Coeff.Neg(&d.Coeff)
+			}
+		}
+		return d
+	}
 
 	// Divide by 10 in a loop. In benchmarks of reduce0.decTest, this is 20%
 	// faster than converting to a string and trimming the 0s from the end.


RaduBerinde:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[decimal.go, line 480 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/20#-Kd6T9aImuHCcwLQKeS2:-Kd6T9aImuHCcwLQKeS3:b-ql2oei) ([raw file](https://github.com/cockroachdb/apd/blob/49c6551c912cb1dfeb810fc6fe27f707a03a0982/decimal.go#L480)):*
> ```Go
> 		i := d.Coeff.Uint64()
> 		e := d.Exponent
> 		for i%10 == 0 {
> ```

maybe add a `for i >= 10000 && i % 10000 == 0 { i /= 10000; e += 4 }`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/20#-:-Kd6UrQcw3dazZQvrou7:bnfp4nl)*

mjibson:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[decimal.go, line 480 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/20#-Kd6T9aImuHCcwLQKeS2:-Kd7Nf2edeVfZQzhKoKQ:bsqq2ce) ([raw file](https://github.com/cockroachdb/apd/blob/49c6551c912cb1dfeb810fc6fe27f707a03a0982/decimal.go#L480)):*

maybe add a `for i >= 10000 && i % 10000 == 0 { i /= 10000; e += 4 }`
</blockquote></details>

Wow, this yielded another very large speed up in the benchmark. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/20)*

RaduBerinde:
*[decimal.go, line 480 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/20#-Kd6T9aImuHCcwLQKeS2:-Kd7gfm7TQO8MtoSN4Q0:b4c9lyr) ([raw file](https://github.com/cockroachdb/apd/blob/49c6551c912cb1dfeb810fc6fe27f707a03a0982/decimal.go#L480)):*

Wow, this yielded another very large speed up in the benchmark. Thanks.
</blockquote></details>

:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/20)*

RaduBerinde:
*[decimal.go, line 480 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/20#-Kd6T9aImuHCcwLQKeS2:-Kd7gtfz_M0wDEzdrVa-:b25l0he) ([raw file](https://github.com/cockroachdb/apd/blob/49c6551c912cb1dfeb810fc6fe27f707a03a0982/decimal.go#L480)):*

:thumbsup: 
</blockquote></details>

The same idea can be useful in the big loop below as well. We could:
 - divide by 10^10 until the remainder is non-zero
 - convert the remainder to `uint64`
 - use uint64 code to count the zeros on the remainder
 - do a final `Quo` with the exact power of 10 (as determined in the previous step). We can precalculate all powers of ten from 1 to 10.

Probably beyond the intention of this PR but maybe file an issue if you think this would help.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/20)*

PR #21, opened 2017-02-16
NumDigits: minor speedups in lookup table logic

mjibson:
Also don't take Abs when not needed.

```
name                old time/op    new time/op    delta
NumDigitsLookup-12     775µs ±14%     381µs ± 3%  -50.80%  (p=0.008 n=5+5)
NumDigitsFull-12      13.0ms ±19%    12.8ms ±11%     ~     (p=1.000 n=5+5)

name                old alloc/op   new alloc/op   delta
NumDigitsLookup-12     115kB ± 0%      22kB ± 0%     ~     (p=0.079 n=4+5)
NumDigitsFull-12      4.49MB ± 0%    3.91MB ± 0%  -12.88%  (p=0.008 n=5+5)

name                old allocs/op  new allocs/op  delta
NumDigitsLookup-12     2.14k ± 0%     0.42k ± 0%  -80.43%  (p=0.008 n=5+5)
NumDigitsFull-12       34.6k ± 0%     31.1k ± 0%  -10.06%  (p=0.008 n=5+5)
```

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/21)

### DIFF
--- a/table.go
+++ b/table.go
@@ -25,13 +25,17 @@ import "math/big"
 // Using this proof, for a given digit count, the map will return the lower number
 // of decimal digits (k) the binary digit count could represent, along with the
 // value of the border between the two decimal digit counts (10^k).
-const digitsTableSize = 128
+const (
+	digitsTableSize = 128
+	digitsTableLen  = digitsTableSize + 1
+)
 
-var digitsLookupTable [digitsTableSize + 1]tableVal
+var digitsLookupTable [digitsTableLen]tableVal
 
 type tableVal struct {
-	digits int64
-	border big.Int
+	digits  int64
+	border  big.Int
+	nborder big.Int
 }
 
 func init() {
@@ -48,14 +52,8 @@ func init() {
 		elem.border.SetInt64(10)
 		curExp.SetInt64(elem.digits)
 		elem.border.Exp(&elem.border, curExp, nil)
+		elem.nborder.Neg(&elem.border)
 	}
-}
-
-func lookupBits(bitLen int) (tableVal, bool) {
-	if bitLen > 0 && bitLen < len(digitsLookupTable) {
-		return digitsLookupTable[bitLen], true
-	}
-	return tableVal{}, false
 }
 
 // NumDigits returns the number of decimal digits of d.Coeff.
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -69,17 +65,25 @@ func NumDigits(b *big.Int) int64 {
...
+		}
+		if a.Cmp(&val.border) < 0 {
 			return val.digits
 		}
 		return val.digits + 1
 	}
 
 	n := int64(float64(bl) / digitsToBitsRatio)
-	a := new(big.Int).Abs(b)
 	e := new(big.Int).Exp(bigTen, big.NewInt(n), nil)
```
nvanbenschoten: Didn't we just add a lookup table for powers of 10 as well?

table_test.go
```diff
@@ -22,6 +22,31 @@ import (
 	"testing"
 )
 
+func BenchmarkNumDigitsLookup(b *testing.B) {
+	b.StopTimer()
+	runTest := func(start string, c byte) {
+		buf := bytes.NewBufferString(start)
+		var offset int
```
nvanbenschoten: Did you mean to use this?

mjibson: Nope, it was for the correctness test. Removed.

table.go
mjibson: Yes, but the lookup table won't ever be used there because we only hit this branch if the lookup table for this function missed. In benchmarks using tableExp10 is 0 faster. I've switched to it to be consistent, but it's not a win.

table.go
```diff
@@ -69,17 +65,25 @@ func NumDigits(b *big.Int) int64 {
...
 		return val.digits + 1
 	}
 
 	n := int64(float64(bl) / digitsToBitsRatio)
-	a := new(big.Int).Abs(b)
-	e := new(big.Int).Exp(bigTen, big.NewInt(n), nil)
+	e := tableExp10(n, nil)
+	a := b
+	if b.Sign() < 0 {
+		a = new(big.Int).Abs(b)
```
nvanbenschoten: If we're going to allocate `a`, why don't we use it as the tmp variable for `tableExp10` as well? 

Also, why does that function say `"The returned value must not be mutated"`? Seems unclear.

table.go
```diff
@@ -69,17 +65,25 @@ func NumDigits(b *big.Int) int64 {
...
 	}
-	if val, ok := lookupBits(bl); ok {
-		ab := new(big.Int).Abs(b)
-		if ab.Cmp(&val.border) < 0 {
+
+	if bl < digitsTableLen {
+		val := digitsLookupTable[bl]
+		a := b
+		if b.Sign() < 0 {
+			a = new(big.Int).Abs(b)
```
nvanbenschoten: Do you think it would be worth adding negative values to the table to avoid this?

table.go
mjibson: Done.

Since tableExp10 may return a value from a lookup table, it should not be mutated. I'm unclear why this is unclear. Can you suggest better wording? For example, this would be wrong:

```
a := tableExp10(5, nil)
a.Add(a, a)
```

because it would modify the big.Int in the lookup table.

table.go
mjibson: Sure. Done. Even more faster and way less allocs! Good suggestion.

table.go
nvanbenschoten: Ah ok, I guess I was confused that the comment talks about what it returns, then talks about what it will set `f` to, then talks about what it returns again. Why not just call `f` `tmpInt` instead and make no guarantee about what it will be set to. This would be a lot more clear, and I doubt we rely on the behavior of `f` being set to `x` anywhere.

nvanbenschoten:
LGTM, although I'd still like to clear up the `tableExp10` question.

nvanbenschoten:
Nice speedups, by the way!

PR #26, opened 2017-02-23
quantize: correct results when rounding

mjibson:
Not sure how we passed the GDA tests before, but the first test case
added in TestQuantize was failing.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/26)

### DIFF
--- a/context.go
+++ b/context.go
@@ -921,9 +921,10 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 	return c.goError(res)
 }
 
-// Quantize sets d to the value of v with x's Exponent.
-func (c *Context) Quantize(d, v, x *Decimal) (Condition, error) {
-	res := c.quantize(d, v, x)
+// Quantize adjusts and rounds v as necessary so it is represented with
+// exponent exp and stores the result in d.
+func (c *Context) Quantize(d, v *Decimal, exp int32) (Condition, error) {
+	res := c.quantize(d, v, exp)
 	if nd := d.NumDigits(); nd > int64(c.Precision) {
 		res |= InvalidOperation
 	}
@@ -931,17 +932,15 @@ func (c *Context) Quantize(d, v, x *Decimal) (Condition, error) {
 	return c.goError(res)
 }
 
-func (c *Context) quantize(d, v, e *Decimal) Condition {
+func (c *Context) quantize(d, v *Decimal, exp int32) Condition {
+	diff := exp - v.Exponent
 	d.Coeff.Set(&v.Coeff)
-	diff := e.Exponent - v.Exponent
 	var res Condition
-	var offset int32
 	if diff < 0 {
 		if diff < MinExponent {
 			return SystemUnderflow | Underflow
 		}
-		e := tableExp10(-int64(diff), nil)
-		d.Coeff.Mul(&d.Coeff, e)
+		d.Coeff.Mul(&d.Coeff, tableExp10(-int64(diff), nil))
 	} else if diff > 0 {
 		p := int32(d.NumDigits()) - diff
 		if p < 0 {
@@ -951,17 +950,36 @@ func (c *Context) quantize(d, v, e *Decimal) Condition {
 			}
 		} else {
 			nc := c.WithPrecision(uint32(p))
+
+			// The idea here is that the resulting d.Exponent after rounding will be 0. We
+			// have a number of, say, 5 digits, but p (our precision) above is set at, say,
+			// 3. So here d.Exponent is set to `-2`. We have a number like `NNN.xx`, where
+			// the `.xx` part will be rounded away. However during rounding of 0.9 to 1.0,
+			// d.Exponent could be set to 1 instead of 0, so we have to reduce it and
+			// increase the coefficient below.
+
+			// Another solution is to set d.Exponent = v.Exponent and adjust it to exp,
+			// instead of setting d.Exponent = -diff and adjusting it to zero. Although
+			// this computes the correct result, it fails the Max/MinExponent checks
+			// during Round and raises underflow flags. Quantize (as per the spec)
+			// is guaranteed to not raise underflow, and using 0 instead of exp as the
+			// target eliminates this problem.
+
+			d.Exponent = -diff
 			// Avoid the c.Precision == 0 check.
-			res = nc.Rounding.Round(nc, d, d)
-			offset = d.Exponent - diff
+			res = nc.rounding().Round(nc, d, d)
+			// Adjust for 0.9 -> 1.0 rollover.
+			if d.Exponent > 0 {
+				d.Coeff.Mul(&d.Coeff, bigTen)
+			}
 		}
 	}
-	d.Exponent = e.Exponent + offset
+	d.Exponent = exp
 	return res
 }
 
 func (c *Context) toIntegral(d, x *Decimal) Condition {
-	res := c.quantize(d, x, decimalOne)
+	res := c.quantize(d, x, 0)
 	// TODO(mjibson): trim here, once trim is in
 	return res
 }
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -421,3 +421,48 @@ func TestContextSetStringt(t *testing.T) {
 		})
 	}
 }
+
+func TestQuantize(t *testing.T) {
+	tests := []struct {
+		s      string
+		e      int32
+		expect string
+	}{
+		{
+			s:      "1.00",
+			e:      -1,
+			expect: "1.0",
+		},
+		{
+			s:      "2.0",
+			e:      -1,
+			expect: "2.0",
+		},
+		{
+			s:      "3",
+			e:      -1,
+			expect: "3.0",
+		},
+		{
+			s:      "9.9999",
+			e:      -2,
+			expect: "10.00",
+		},
+	}
+	c := BaseContext.WithPrecision(10)
+	for _, tc := range tests {
+		t.Run(fmt.Sprintf("%s: %d", tc.s, tc.e), func(t *testing.T) {
+			d, _, err := NewFromString(tc.s)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if _, err := c.Quantize(d, d, tc.e); err != nil {
+				t.Fatal(err)
+			}
+			s := d.String()
+			if s != tc.expect {
+				t.Fatalf("expected: %s, got: %s", tc.expect, s)
+			}
+		})
+	}
+}
--- a/error.go
+++ b/error.go
@@ -124,9 +124,15 @@ func (e *ErrDecimal) Pow(d, x, y *Decimal) *Decimal {
 	return e.op3(d, x, y, e.Ctx.Pow)
 }
 
-// Quantize performs e.Ctx.Quantize(d, v, x) and returns d.
-func (e *ErrDecimal) Quantize(d, v, x *Decimal) *Decimal {
-	return e.op3(d, v, x, e.Ctx.Quantize)
+// Quantize performs e.Ctx.Quantize(d, v, exp) and returns d.
+func (e *ErrDecimal) Quantize(d, v *Decimal, exp int32) *Decimal {
+	if e.Err() != nil {
+		return d
+	}
+	res, err := e.Ctx.Quantize(d, v, exp)
+	e.Flags |= res
+	e.err = err
+	return d
 }
 
 // Quo performs e.Ctx.Quo(d, x, y) and returns d.
--- a/gda_test.go
+++ b/gda_test.go
@@ -271,7 +271,7 @@ func (tc TestCase) Run(c *Context, done chan error, d, x, y *Decimal) (res Condi
 	case "power":
 		res, err = c.Pow(d, x, y)
 	case "quantize":
-		res, err = c.Quantize(d, x, y)
+		res, err = c.Quantize(d, x, y.Exponent)
 	case "reduce":
 		res, err = c.Reduce(d, x)
 	case "remainder":


RaduBerinde:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[context.go, line 934 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlYwMKMII3mINg1bhx:-KdlYwMKMII3mINg1bhy:b1xm4k) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L934)):*
> ```Go
> }
> 
> func (c *Context) quantize(d, v, e *Decimal) Condition {
> ```

add a comment explaining what this does. Also why do we need `e` to be a `*Decimal`? Seems like we could just pass the exponent as an int32 or whatever

---

*[context.go, line 955 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlZtLlV4M7RHJzA263:-KdlZtLmdZ1UXb2hO6Aj:bpcdkjp) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L955)):*
> ```Go
> 			nc := c.WithPrecision(uint32(p))
> 			neg := d.Sign() < 0
> 			d.Exponent = -diff
> ```

`d.Exponent` here is effectively just a temporary variable which just makes things confusing.

---

*[context.go, line 961 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-Kdl_Kmz13BheEJ7cS7y:-Kdl_Kn-5e-RvNtMDh6g:bc92ts8) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L961)):*
> ```Go
> 			for d.Exponent > 0 {
> 				d.Exponent--
> 				d.Coeff.Mul(&d.Coeff, bigTen)
> ```

I'm a little confused why we want to do this. Isn't 100e-4 equivalent to 1e-2? And isn't the latter preferred for efficiency?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

mjibson:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[context.go, line 934 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlYwMKMII3mINg1bhx:-Kdm3tZVj-6oYM241c0z:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L934)):*

add a comment explaining what this does. Also why do we need `e` to be a `*Decimal`? Seems like we could just pass the exponent as an int32 or whatever
</blockquote></details>

Done.

---

*[context.go, line 955 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlZtLlV4M7RHJzA263:-Kdm3vdOM8mQ0Dl5tvyP:b8a80h4) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L955)):*

`d.Exponent` here is effectively just a temporary variable which just makes things confusing.
</blockquote></details>

The idea here is that the resulting d.Exponent after rounding will be 0. This is because we have a number of, say, 5 digits, but p (our precision) above is set a 3. So here d.Exponent is set to `-2`. We have a number like `NNN.xx`, where the `.xx` part will be rounded away. However during rounding of 0.9 to 1.0, d.Exponent will be set to 1 instead of 0, so we have to reduce it and increase the coefficient below. It's not quite a temporary variable.

I've added this comment to try to explain my reasoning. (However it's possible my reasoning or implementation is incorrect.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

RaduBerinde:

---

*[context.go, line 955 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlZtLlV4M7RHJzA263:-KdmEYWeZ7ecAnNsoLBR:bjfuetr) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L955)):*

The idea here is that the resulting d.Exponent after rounding will be 0. This is because we have a number of, say, 5 digits, but p (our precision) above is set a 3. So here d.Exponent is set to `-2`. We have a number like `NNN.xx`, where the `.xx` part will be rounded away. However during rounding of 0.9 to 1.0, d.Exponent will be set to 1 instead of 0, so we have to reduce it and increase the coefficient below. It's not quite a temporary variable.

I've added this comment to try to explain my reasoning. (However it's possible my reasoning or implementation is incorrect.)
</blockquote></details>

Ah. I missed the fact that `Round` uses and changes `d.Exponent`.

What I don't understand then is why we are setting the exponent to `-diff` and then trying to make it 0 instead of letting it be what it is, rounding to the calculated precision, and then adjusting to make sure it's `exp`? (I realize these are equivalent, but for clarity)

---

*[context.go, line 961 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-Kdl_Kmz13BheEJ7cS7y:-KdmGCyzHhrihthTysWN:b-shavap) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L961)):*

I'm a little confused why we want to do this. Isn't 100e-4 equivalent to 1e-2? And isn't the latter preferred for efficiency?
</blockquote></details>

This loop could have a lot of iterations no? e.g. for 0.99999999 -> 1.0 Why not mul with `tableExp10` like above? Better yet - move this whole branch to the top, and let the other branch handle this, something like:
```
if d.Exponent > exp {
   // round (which brings the exponent to at most `exp`).
}
if d.Exponent < exp {
  ..
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

RaduBerinde:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[context.go, line 925 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdmITt1mKUEwIkDEOwu:-KdmITt1mKUEwIkDEOwv:b-nqcavb) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L925)):*
> ```Go
> 
> // Quantize sets d to the value of v with x's Exponent.
> func (c *Context) Quantize(d, v, x *Decimal) (Condition, error) {
> ```

can't we change this to take an exponent too?

---

*[context.go, line 934 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdmIWSF-f1pF1a7_jNP:-KdmIWSF-f1pF1a7_jNQ:bj2xgw5) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L934)):*
> ```Go
> }
> 
> // quantize sets d to the value of v with exponent exp.
> ```

this description is still not 100% clear.. maybe "adjusts and rounds v as necessary so it is represented with exponent exp and stores the result in `d` (same above). It's fine if we just comment the function above (I didn't notice it earlier in reviewboard)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

nvanbenschoten:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[context.go, line 925 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdmITt1mKUEwIkDEOwu:-KdnnL5Ucoe4efbA1EYX:bkw2pie) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L925)):*

can't we change this to take an exponent too?
</blockquote></details>

+1 if this restrictive interface ever forces us to allocate unnecessarily.

---

*[context.go, line 973 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdnpRtvRaCaPCRx1SuW:-KdnpRtvRaCaPCRx1SuX:b-vqxooy) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L973)):*
> ```Go
> 			// unexpectedly removes a negative sign when converting from -9 to -10. This
> 			// check is needed until it is fixed.
> 			if neg && d.Sign() > 0 {
> ```

Can we log something here or panic? If there's a bug, it would be nice to catch it ASAP.

---

*[decimal_test.go, line 428 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdnoXrrXTIgZhXCF3Xs:-KdnoXrsuX1wuIOC3MTX:b-fmz4my) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/decimal_test.go#L428)):*
> ```Go
> 	tests := []struct {
> 		s      string
> 		e      int32
> ```

no tests for `e > 0` or `e == 0`? Also, no tests for `|e| >= len(s)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

mjibson:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[context.go, line 955 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdlZtLlV4M7RHJzA263:-KdpHeCrig6klzKOqO2S:b-ipjfl4) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L955)):*

Ah. I missed the fact that `Round` uses and changes `d.Exponent`.

What I don't understand then is why we are setting the exponent to `-diff` and then trying to make it 0 instead of letting it be what it is, rounding to the calculated precision, and then adjusting to make sure it's `exp`? (I realize these are equivalent, but for clarity)
</blockquote></details>

I've added another comment about why we do it this way instead of the way you describe. It's because the way you describe raises Underflow flags and basing around 0 doesn't, and the Quantize spec says it doesn't raise Underflow. Specifically, it fails the quax545 test (https://github.com/cockroachdb/apd/blob/master/testdata/quantize.decTest#L613) with incorrect flags `expected flags "inexact"; got flags "underflow, inexact, subnormal, clamped"`.

---

*[context.go, line 961 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-Kdl_Kmz13BheEJ7cS7y:-KdpIDaHtrZxfB9Uv-gZ:b-ucoi1x) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L961)):*

This loop could have a lot of iterations no? e.g. for 0.99999999 -> 1.0 Why not mul with `tableExp10` like above? Better yet - move this whole branch to the top, and let the other branch handle this, something like:
```
if d.Exponent > exp {
   // round (which brings the exponent to at most `exp`).
}
if d.Exponent < exp {
  ..
}
```
</blockquote></details>

The api for this function is that the resulting `d.Exponent == exp`, so we must adjust d.Coeff to be whatever it needs to in order to get that. Adding zeros to the coeff is the only way to do that.

The 0.99999 -> 1.0 case will still only loop one time in the for loop. Although it was a for loop, it should only execute at most 1 times. I've changed it to an if statement, and the tests still pass.

---

*[context.go, line 925 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdmITt1mKUEwIkDEOwu:-KdpFCeHsUZ8-PSk6vV7:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L925)):*

+1 if this restrictive interface ever forces us to allocate unnecessarily.
</blockquote></details>

Done.

---

*[context.go, line 934 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdmIWSF-f1pF1a7_jNP:-KdpFgy-jQd1uPjYBhze:b-5fj118) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L934)):*

this description is still not 100% clear.. maybe "adjusts and rounds v as necessary so it is represented with exponent exp and stores the result in `d` (same above). It's fine if we just comment the function above (I didn't notice it earlier in reviewboard)
</blockquote></details>

Done. This function is indeed difficult to describe.

---

*[context.go, line 973 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdnpRtvRaCaPCRx1SuW:-KdpMp3EnYbh-9AYyIgF:bxxdsgb) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/context.go#L973)):*

Can we log something here or panic? If there's a bug, it would be nice to catch it ASAP.
</blockquote></details>

I have another PR out to just fix this bug.

---

*[decimal_test.go, line 428 at r2](https://reviewable.io:443/reviews/cockroachdb/apd/26#-KdnoXrrXTIgZhXCF3Xs:-KdpJKGqnk_TW9rCanI3:bb0xb8l) ([raw file](https://github.com/cockroachdb/apd/blob/1e19da4d32b9f78ab87f60e290e7273788161251/decimal_test.go#L428)):*

no tests for `e > 0` or `e == 0`? Also, no tests for `|e| >= len(s)`?
</blockquote></details>

The tests here are just a supplement to the GDA tests, and not meant to be exhaustive. I found a single test case that failed and wasn't also covered by GDA (the 1.00 test). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26)*

RaduBerinde:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[context.go, line 961 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/26#-Kdl_Kmz13BheEJ7cS7y:-KdphCBiUr7RL7Kcyc3G:b-wcpdx4) ([raw file](https://github.com/cockroachdb/apd/blob/190244f232f176a63c917a5986c65ff9ef4d9020/context.go#L961)):*

The api for this function is that the resulting `d.Exponent == exp`, so we must adjust d.Coeff to be whatever it needs to in order to get that. Adding zeros to the coeff is the only way to do that.

The 0.99999 -> 1.0 case will still only loop one time in the for loop. Although it was a for loop, it should only execute at most 1 times. I've changed it to an if statement, and the tests still pass.
</blockquote></details>

Weird, I would have expected that rounding 0.99999 would have given us 1 with exponent 0. But I see that this is working. Since we now have an `if`, can we add a check with panic that the exponent is 1? (just so it's clear)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26#-:-KdpjXteSH10b_UlD2Gr:bnfp4nl)*

nvanbenschoten:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/26#-:-KdsIz0XmsTdv_Bshg6p:bnfp4nl)*

PR #27, opened 2017-02-25
round: fix rounding bug found by quantize

mjibson:
When rounding from -0.9 to -1.0, the coefficient would be first set
to 0, and then 1 added to it. Since 0 isn't negative, 1 was added
instead of subtracted as it should have been. The fix is to record
the sign before rounding.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/27)

### DIFF
--- a/context.go
+++ b/context.go
@@ -228,7 +228,7 @@ func (c *Context) Quo(d, x, y *Decimal) (Condition, error) {
 			half := dividend.Cmp(divisor)
 			rounding := c.rounding()
 			if rounding(&quo.Coeff, half) {
-				roundAddOne(&quo.Coeff, &diff)
+				roundAddOne(&quo.Coeff, &diff, 1 /* positive */)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


nvanbenschoten:
:lgtm:

---

---

*[context.go, line 231 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/27#-KdsJQ0-jLYsCfP3DWyj:-KdsJQ0-jLYsCfP3DWyk:b-k91aan) ([raw file](https://github.com/cockroachdb/apd/blob/49f122b8e814b86526845d19dc96f4b57747088c/context.go#L231)):*
> ```Go
> 			rounding := c.rounding()
> 			if rounding(&quo.Coeff, half) {
> 				roundAddOne(&quo.Coeff, &diff, 1)
> ```

consider either:

```
const positive = 1
roundAddOne(&quo.Coeff, &diff, positive)
```

or 

```
roundAddOne(&quo.Coeff, &diff, 1 /* positive */)
```

---

*[round.go, line 94 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/27#-KdsJGnTpdVHd4i9-dTl:-KdsJGnTpdVHd4i9-dTm:bhioiyu) ([raw file](https://github.com/cockroachdb/apd/blob/49f122b8e814b86526845d19dc96f4b57747088c/round.go#L94)):*
> ```Go
> }
> 
> // roundAddOne adds 1 to abs(b).
> ```

mention the sign parameter in this comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/27#-:-KdsJg2ezcacSvK6a549:bnfp4nl)*

PR #30, opened 2017-03-08
pow: correctly compute exactness

mjibson:
Power calculations of x**y where y is an integer will now correctly
be flagged as exact when possible. This avoids lots of work and is
thus much faster for these operations.

Remove some unneeded precision changing in integerPower. It
was leftover from the inf code which did bad things with 1/x
calculations. This was causing some overflow problems that shouldn't
have been present.

This change reduces `go test` time from 2.5s to 0.9s.

```
name          old time/op  new time/op  delta
GDA/power-12   666ms ±16%   387ms ± 6%  -41.86%  (p=0.008 n=5+5)
```

---

First commit is from another PR, but has a fix this needs.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/30)

### DIFF
--- a/context.go
+++ b/context.go
@@ -842,36 +842,20 @@ func (c *Context) integerPower(d, x *Decimal, y *big.Int) (Condition, error) {
 	}
 
 	if neg {
-		e := z.Exponent
-		if e < 0 {
-			e = -e
-		}
-		qc := c.WithPrecision((uint32(z.NumDigits()) + uint32(e)) * 2)
-		ed.Ctx = qc
 		ed.Quo(z, decimalOne, z)
-		ed.Ctx = c
 	}
 	return ed.Flags, ed.Err()
 }
 
 // Pow sets d = x**y.
 func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
-	// x ** 1 == x
-	if y.Cmp(decimalOne) == 0 {
-		return c.Round(d, x)
-	}
-	// 1 ** x == 1
-	if x.Cmp(decimalOne) == 0 {
-		return c.Round(d, x)
-	}
-
 	// Check if y is of type int.
 	tmp := new(Decimal)
 	if _, err := c.Abs(tmp, y); err != nil {
 		return 0, errors.Wrap(err, "Abs")
 	}
 	integ, frac := new(Decimal), new(Decimal)
-	tmp.Modf(integ, frac)
+	y.Modf(integ, frac)
 	yIsInt := frac.Sign() == 0
 
 	xs := x.Sign()
@@ -910,6 +894,19 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 	p += 4
 
 	nc := BaseContext.WithPrecision(p)
+
+	if yIsInt {
+		// If integ.Exponent > 0, we need to add trailing 0s to integ.Coeff.
+		res := c.quantize(integ, integ, 0)
+		nres, err := nc.integerPower(d, x, &integ.Coeff)
+		res |= nres
+		if err != nil {
+			return res, err
+		}
+		res |= c.round(d, d)
+		return c.goError(res)
+	}
+
 	ed := MakeErrDecimal(nc)
 
 	ed.Abs(tmp, x)
@@ -925,9 +922,7 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 		return ed.Flags, err
 	}
 	res := c.round(d, tmp)
-	if !yIsInt {
-		res |= Inexact
-	}
+	res |= Inexact
 	return c.goError(res)
 }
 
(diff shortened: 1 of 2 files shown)


eisenstatdavid:
:lgtm: pending rebase, but I'm wondering if it would make sense to use integerPower on the integer part and do the log/mul/exp calculation on the fractional part and multiply. That implementation would be obviously monotone, which is worth something, and it might be faster (though I really have no idea; if it isn't, does this PR slow some cases with large exponents down?)

---

Reviewed 2 of 2 files at r1.
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 864 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/30#-KeiY99FOcwHdMhaGCYS:-KeiY99FOcwHdMhaGCYT:b-yzfqfd) ([raw file](https://github.com/cockroachdb/apd/blob/7884ea1a6ef6ecf37ed0bc57d490b61d0bec026c/context.go#L864)):*
> ```Go
> 	}
> 	// 1 ** x == 1
> 	if x.Cmp(decimalOne) == 0 {
> ```

This case isn't handled specially by the new code, is it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-KeiZcqxF1hI9wVKHzFG:b-s7m4h)*

mjibson:
I can't figure out how to split up and recombine the integer and fractional parts to test if this is faster. Is there an identity I'm missing?

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 864 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/30#-KeiY99FOcwHdMhaGCYS:-KejCzc0S2EPozha_mRK:b-c42e9w) ([raw file](https://github.com/cockroachdb/apd/blob/7884ea1a6ef6ecf37ed0bc57d490b61d0bec026c/context.go#L864)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

This case isn't handled specially by the new code, is it?
</blockquote></details>

No, but it doesn't need to be. It was handled specially as a performance improvement, but as a side effect it incorrectly flagged exactness. Pow is fast enough now that we don't need to special case it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-Kej_PO8RV-l0rJDuEQ-:br08oit)*

eisenstatdavid:
pow(x, y) = pow(x, integ(y)) * pow(x, frac(y))?

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-Kejb6Vhm7uJn_cXZKrr:b8lfa3c)*

mjibson:
I don't think this PR slows down integer cases with large exponents. If the exponents are very large, then they will quickly overflow anyway, so I've been having difficulty constructing an interesting test case.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-KejayQQCc98rt04esPB:bd95mn4)*

mjibson:
Oh duh. I was thinking about separating x into integ and frac parts. Not thinking clearly.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-KejbOTDHvN5-ha4fMot:b-fxmzon)*

eisenstatdavid:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[context.go, line 864 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/30#-KeiY99FOcwHdMhaGCYS:-KejcMQLEOLCzZpD9edP:bksy8p7) ([raw file](https://github.com/cockroachdb/apd/blob/7884ea1a6ef6ecf37ed0bc57d490b61d0bec026c/context.go#L864)):*

No, but it doesn't need to be. It was handled specially as a performance improvement, but as a side effect it incorrectly flagged exactness. Pow is fast enough now that we don't need to special case it.
</blockquote></details>

Weird omission from the spec. OK.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30)*

mjibson:
You were right. Separating the integer and fractional calculations out yielded another 10% speed up. Added a second commit with this. RFAL.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-KejeLO8uF1XXIWmqrfQ:b-grjfxw)*

eisenstatdavid:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/30#-:-KejhqTG7i7A5iqva-ih:bnfp4nl)*

PR #33, opened 2017-03-16
Add NaN and Inf support; allow result-as-operand

mjibson:
This is a fairly big PR and incorporates two different features. I can break it up into 2 PRs if you prefer, but it's at least different commits. No rush on this since it's pretty big. Also I'm happy to find another reviewer if you'd prefer to not do it.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/33)


eisenstatdavid:
I started reviewing today. No need to split, but yeah, this review will take a while.

eisenstatdavid:
This PR is in good shape despite the many comments below.

On subject of testing, I haven't looked at the IBM tests carefully (and the exclusions) because I know that they're pretty comprehensive, but do you know what kind of branch coverage they achieve on this code?

---

Reviewed 8 of 8 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 29 unresolved discussions.

---

*[const.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMmkLJ9g-XF23V-LJb:-KfMmkLJ9g-XF23V-LJc:bklgfsk) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/const.go#L79)):*
> ```Go
> 			MaxExponent: MaxExponent,
> 			MinExponent: MinExponent,
> 		}
> ```

Could there be other places that need this fix?

---

*[context.go, line 125 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMnKLlWi9wjmkKyeaq:-KfMnHqM6qunpk-68YwD:bjmemz5) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L125)):*
> ```Go
> 	}
> 	xn := x.Negative
> 	yn := y.Negative
> ```

Maybe unify the branches below with `yn := y.Negative != subtract`? The final result would be sth like

```
xn := x.Negative
yn := y.Negative != subtract  // XOR
if xi, yi := x.Form == Infinite, y.Form == Infinite; xi || yi {
  if xi && yi && xn != yn {
    d.Set(decimalNaN)
    return c.goError(InvalidOperation)
  } else if xi {
    d.Set(x)
  } else {
    d.Set(decimalInfinity)
    d.Negative = yn
  }
  return 0, nil
}
```

---

*[context.go, line 173 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMvRBCuCUr5i60vpFe:-KfMvRBDdZLfWiuKCWPO:bttezp9) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L173)):*
> ```Go
> 	if subtract {
> 		yn = !yn
> 	}
> ```

This would be subsumed by `yn := y.Negative != subtract` above.

---

*[context.go, line 175 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMvF5HcEcFY3HE9fPL:-KfMvF5HcEcFY3HE9fPM:bnewko) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L175)):*
> ```Go
> 	}
> 	subtract = xn != yn
> 	if subtract {
> ```

Since `subtract` is never used again, perhaps change the test to `xn == yn` (and reverse the branches).

For this branch, atomically with the below suggestion, suggest

```
if xn {
  d.Coeff.Sub(b, a)
} else {
  d.Coeff.Sub(a, b)
}
// special case for RoundFloor on the following line if necessary
d.Negative = d.Coeff.Sign() < 0
d.Coeff.Abs(&d.Coeff)
```

---

*[context.go, line 179 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMyyAsIuo029qu-9Zf:-KfMyyAteRFU3uSNxsdj:bk1p5gb) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L179)):*
> ```Go
> 		d.Coeff.Abs(&d.Coeff)
> 	} else {
> 		d.Coeff.Add(a, b)
> ```

Atomically with the above suggestion, suggest `d.Negative = xn` here. This logic covers negative zero + negative zero and negative zero - positive zero.

---

*[context.go, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMvbbxVjHoaQIiqDy2:-KfMuzDzq0drmu548L5W:bpixust) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L189)):*
> ```Go
> 		}
> 	} else {
> 		d.Negative = xn && yn
> ```

`&& x.Coeff.Sign() == 0 && y.Coeff.Sign() == 0` unless the spec has a gratuitous difference from IEEE (but see above). Is this case tested?

---

*[context.go, line 232 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfN1SorG8GpnSvOIXre:-KfN1SosmA_xtaFtSD_8:b-h8aviy) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L232)):*
> ```Go
> 		return res, err
> 	}
> 	if xi, yi := x.Form == Infinite, y.Form == Infinite; xi || yi {
> ```

This logic could be simplified to

```
if x.Form == Infinite || y.Form == Infinite {
  if x.IsZero() || y.IsZero() {
    d.Set(decimalNaN)
    res = InvalidOperation
  } else {
    d.Set(decimalInfinity)
    d.Negative = x.Negative != y.Negative
  }
  return c.goError(res)
}
```

Definitely at least suppress the special case for `xi && yi`.

---

*[context.go, line 280 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfN90DxAJrLbSd27tWu:-KfN90Dy1g78z-lQcKrG:b9kkkf7) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L280)):*
> ```Go
> 		} else if xi {
> 			d.Set(decimalInfinity)
> 			d.Negative = x.Negative != y.Negative
> ```

Don't we need to check whether `y` is zero and set a flag if so? Should all of this added code come before the precision check?

---

*[context.go, line 283 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNCKMb8fcuQNkzi1cP:-KfNCKMb8fcuQNkzi1cQ:bsxaum7) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L283)):*
> ```Go
> 		} else {
> 			d.SetInt64(0)
> 			d.Exponent = c.etiny()
> ```

Idle question: how does setting a very small exponent interact with `upscale`?

---

*[context.go, line 410 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNM6xIABVwYHq8e4wX:-KfNM6xIABVwYHq8e4wY:bdmp4me) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L410)):*
> ```Go
> 	}
> 	var res Condition
> 	if xi, yi := x.Form == Infinite, y.Form == Infinite; xi || yi {
> ```

Could the duplicate logic here be pulled out as (e.g.) `quoSpecials`? Maybe there's another way to reduce the code duplication across division routines.

---

*[context.go, line 441 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNN2H1-OB-ZbFmt7gI:-KfNN2H1-OB-ZbFmt7gJ:b-jfa2gu) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L441)):*
> ```Go
> 	d.Form = Finite
> 	if d.NumDigits() > int64(c.Precision) {
> 		d.Set(decimalNaN)
> ```

Why NaN instead of rounding? The spec seems silent on this issue.

---

*[context.go, line 462 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNOFKJNK5kzXEfo8c5:-KfNOFKJNK5kzXEfo8c6:b-a0l1b3) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L462)):*
> ```Go
> 	}
> 	if y.Form == Infinite {
> 		d.Set(x)
> ```

Don't we need to invert the sign when y is negative? IIRC, floating point division should satisfy -x / y = x / -y = -(x / y).

---

*[context.go, line 617 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQ1pXL8lG8m_oeko3:-KfNQ1pXL8lG8m_oeko4:btrv4vj) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L617)):*
> ```Go
> 	//     x_{n+1} = 1/3 * ( 2 * x_n + (d / x_n / x_n) ).
> 
> 	if set, res, err := c.rootSpecials(d, x); set {
> ```

Does this `Cbrt` return NaN on negative inputs? Should it?

---

*[context.go, line 891 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQOAkhPNeqWCb5Lkf:-KfNQOAkhPNeqWCb5Lkg:b-vyfczi) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L891)):*
> ```Go
> 		return res, err
> 	}
> 	if x.Sign() < 0 {
> ```

`logSpecials`?

---

*[context.go, line 934 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNR3jlRcI5Gz3s9MMf:-KfNR3jlRcI5Gz3s9MMg:b-hbpg2c) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L934)):*
> ```Go
> 	// Transactions on Mathematical Software, Vol 12 #2, pp79-91, ACM, June 1986.
> 
> 	if set, res, err := c.setIfNaN(d, x); set {
> ```

How many of these functions need to set the Form field? I guess the transcendental functions don't because they're computed by iterative algorithms that use lower-level functions.

---

*[context.go, line 1214 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg78nhYXK447Gk40Uq:-Kfg78nhYXK447Gk40Ur:bgb531) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L1214)):*
> ```Go
> 
> func (c *Context) quantize(d, v *Decimal, exp int32) Condition {
> 	if v.Form == Infinite || exp < c.etiny() {
> ```

Could this check go in `Quantize`? I'm not sure why we would test `d.NumDigits()` when `d` is not a number.

---

*[decimal.go, line 43 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg7xk9PojutApH5n2Y:-Kfg7xk9PojutApH5n2Z:b-4jhqvb) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L43)):*
> ```Go
> 
> const (
> 	// These are in total-ordering order, and so cannot be otherwise sorted.
> ```

I had to read this comment a couple times to get it. Perhaps "`CmpTotal` assumes that the order of these constants reflects the total order on decimals"? `cmpOrder` seems to make further assumptions, which may be worth mentioning.

---

*[decimal.go, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9ByW6akrodhhWEE6:-Kfg9ByXtsIgI_aqI2xv:b-nqke0h) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L114)):*
> ```Go
> 		isNaN = true
> 		d.Form = NaN
> 		s = s[3:]
> ```

Rather than count here, perhaps introduce a function like

```
func ConsumePrefix(s, prefix string) (string, bool) {
  if strings.HasPrefix(s, prefix) {
    return s[len(prefix):], true
  }
  return s, false
}
```

and use throughout.

---

*[decimal.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9lghL5ky0eai33el:-Kfg9lghL5ky0eai33em:b-1vxlrz) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L132)):*
> ```Go
> 	}
> 
> 	// Until there are no parse errors, leave as NaN.
> ```

Couldn't this be done earlier?

---

*[decimal.go, line 136 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9yS5TfQ9qMEbjwRh:-Kfg9yS5TfQ9qMEbjwRi:b-9ozdn4) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L136)):*
> ```Go
> 
> 	var exps []int64
> 	if i := strings.IndexAny(s, "eE"); i >= 0 {
> ```

We're only looking for lowercase e now, so we can use `IndexByte`.

---

*[decimal.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgAPHyXrFdam38VzGo:-KfgAPHzkLjCQvSoRA2T:bkji12h) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L244)):*
> ```Go
> 	var s string
> 
> 	switch d.Form {
> ```

Is it worth pulling out the duplicate handling for specials and negative numbers between this function and the previous function?

---

*[decimal.go, line 511 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgCFXYKSHFaqrmtXAi:-KfgCFXZ4QzsQE-jvCbd:bk6hktu) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L511)):*
> ```Go
> 	switch d.Form {
> 	case Finite:
> 		// d and x have the same sign and form, compare their value.
> ```

Unless I'm mistaken, `cmpOrder` returns `0` for all finite decimals, both negative and positive, so `d` and `x` don't necessarily have the same sign. I guess this works anyway because `Cmp` handles the signs too. See below.

---

*[decimal.go, line 536 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgCRLK6dcFFEOzuMuz:-KfgCRLLY_GGVwhm3SHV:b-fqpaq6) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L536)):*
> ```Go
> 	v := int(d.Form)
> 	if d.Negative {
> 		v = -v
> ```

Probably `v = ^v` (or `v = -(v + 1)` if that's too clever); see above.

---

*[decimal.go, line 568 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgF3-W9x8cgS66XbPm:-KfgF3-Xp8pIoweH9ie1:b-h7lggo) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L568)):*
> ```Go
> 	ds := d.Sign()
> 	xs := x.Sign()
> 	gt := 1
> ```

Wait to define these until after checking the signs and leave a comment explaining them. It looks strange at first that the sign tests use ±1 while the rest uses the "constants".

---

*[decimal.go, line 620 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgFOLAvhU8jxyBfgJj:-KfgFLXaUgJ1NZwdbV6y:b-1md5xl) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L620)):*
> ```Go
> 	}
> 
> 	funcs := []func() bool{
> ```

I'm confused by the inner functions. We never seem to set `v` and then not return true, so couldn't we just return from the outer function instead?

---

*[decimal.go, line 666 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgFsONXJz4UINSDtWy:-KfgFsONXJz4UINSDtWz:bkgg5eg) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L666)):*
> ```Go
> //	+1 if d.Negative == false
> //
> func (d *Decimal) Sign() int {
> ```

I prefer

```
if d.Form == Finite && d.Coeff.Sign() == 0 {
  return 0
}
if d.Negative {
  return -1
}
return 1
```

---

*[decimal.go, line 686 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgG8VuUyGlgP60Sq0r:-KfgG8VuUyGlgP60Sq0s:b-cem55o) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L686)):*
> ```Go
> // IsZero returns true if d == 0 or -0.
> func (d *Decimal) IsZero() bool {
> 	return d.Sign() == 0 && d.Form == Finite
> ```

Is the Form check necessary?

---

*[gda_test.go, line 449 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgHLQ_Vd7EHlnzxI0V:-KfgHLQaCXZ6d5Jr1tB-:bojswil) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/gda_test.go#L449)):*
> ```Go
> 					t.Skip("x ** large y")
> 				}
> 			case "quantize":
> ```

Doesn't `Quantize` handle infinite operands now?

---

*[gda_test.go, line 627 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgHjqkQQIrLQt6NN5k:-KfgHjqkQQIrLQt6NN5l:b-loxd9r) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/gda_test.go#L627)):*
> ```Go
> 					tc.Result = "sNaN"
> 				}
> 				if strings.ToUpper(s) != strings.ToUpper(tc.Result) {
> ```

`strings.EqualFold`? Kind of pedantic here admittedly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/33#-:-KfgLgWaD6QUewBO92N_:bevlt4b)*

mjibson:
I ran `go test -coverprofile` which had a 93.6% test coverage. Most of the non-coverage was for error handling lines that were never triggered and some string and other misc stuff that I'll add soon. But all of the meat in decimal.go and context.go appears to be covered. The only thing not covered are two branches in Sqrt that attempt to adjust for off-by-one errors. These will be investigated.

---

Review status: 6 of 8 files reviewed at latest revision, 29 unresolved discussions.

---

*[const.go, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMmkLJ9g-XF23V-LJb:-KfgtCZhslLzosIgFN8K:b-3e8f6q) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/const.go#L79)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Could there be other places that need this fix?
</blockquote></details>

I can't think of any.

---

*[context.go, line 125 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMnKLlWi9wjmkKyeaq:-KfguyYegQs4LWcLF8kz:be06y3h) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L125)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Maybe unify the branches below with `yn := y.Negative != subtract`? The final result would be sth like

```
xn := x.Negative
yn := y.Negative != subtract  // XOR
if xi, yi := x.Form == Infinite, y.Form == Infinite; xi || yi {
  if xi && yi && xn != yn {
    d.Set(decimalNaN)
    return c.goError(InvalidOperation)
  } else if xi {
    d.Set(x)
  } else {
    d.Set(decimalInfinity)
    d.Negative = yn
  }
  return 0, nil
}
```
</blockquote></details>

I spent a few minutes trying to get this to work (that is, all of your comments about this function) but failed, as some small subset of tests always failed. I agree it seems like there's a simpler solution than what we have here but I currently think it should be done in another PR. I've created #37 to track this.

---

*[context.go, line 175 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMvF5HcEcFY3HE9fPL:-KfgzOUgwGJ_cU9UBAZ4:b-xwlxoe) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L175)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Since `subtract` is never used again, perhaps change the test to `xn == yn` (and reverse the branches).

For this branch, atomically with the below suggestion, suggest

```
if xn {
  d.Coeff.Sub(b, a)
} else {
  d.Coeff.Sub(a, b)
}
// special case for RoundFloor on the following line if necessary
d.Negative = d.Coeff.Sign() < 0
d.Coeff.Abs(&d.Coeff)
```
</blockquote></details>

I've made the `xn == yn` change, but not larger one.

---

*[context.go, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMvbbxVjHoaQIiqDy2:-Kfgy5cwnufjEV3-xIuh:b-qjcn4h) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L189)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

`&& x.Coeff.Sign() == 0 && y.Coeff.Sign() == 0` unless the spec has a gratuitous difference from IEEE (but see above). Is this case tested?
</blockquote></details>

Yes, this branch is covered during testing.

---

*[context.go, line 232 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfN1SorG8GpnSvOIXre:-Kfgz6tIAdfUvJmEf-qN:b-mtc4te) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L232)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

This logic could be simplified to

```
if x.Form == Infinite || y.Form == Infinite {
  if x.IsZero() || y.IsZero() {
    d.Set(decimalNaN)
    res = InvalidOperation
  } else {
    d.Set(decimalInfinity)
    d.Negative = x.Negative != y.Negative
  }
  return c.goError(res)
}
```

Definitely at least suppress the special case for `xi && yi`.
</blockquote></details>

This change fails 18 test cases.

---

*[context.go, line 280 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfN90DxAJrLbSd27tWu:-Kfgzy0oclbmxko-9wQW:b-ac31ij) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L280)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Don't we need to check whether `y` is zero and set a flag if so? Should all of this added code come before the precision check?
</blockquote></details>

I don't think we need a `y == 0` check here because it's covered in the tests. For example (https://github.com/cockroachdb/apd/blob/c7ade124510c862bab38aa30f412aebe6c7c2abd/testdata/divide.decTest#L597):

```
divx784 divide  Inf   0     ->  Infinity
```

(This is `Inf / 0 == Inf`)

The precision check was added by me to prevent footgun problems. I think that this block should stay above the precision check because in these cases we can produce correct, fast results.

---

*[context.go, line 283 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNCKMb8fcuQNkzi1cP:-KfgzXQLgAYDfKuqqUZg:b-anbdtj) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L283)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Idle question: how does setting a very small exponent interact with `upscale`?
</blockquote></details>

upscale returns an error if the differing exponents are too different: https://github.com/cockroachdb/apd/blob/c7ade124510c862bab38aa30f412aebe6c7c2abd/decimal.go#L350

---

*[context.go, line 410 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNM6xIABVwYHq8e4wX:-Kfh1Q_cmlB5YNkaPMuy:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L410)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Could the duplicate logic here be pulled out as (e.g.) `quoSpecials`? Maybe there's another way to reduce the code duplication across division routines.
</blockquote></details>

Done.

---

*[context.go, line 441 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNN2H1-OB-ZbFmt7gI:-Kfh1ar8m5V_U9jemV5j:bfoxpzh) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L441)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Why NaN instead of rounding? The spec seems silent on this issue.
</blockquote></details>

The spec says: "The exponent of the result must be 0. Hence, if the result cannot be expressed exactly within precision digits, the operation is in error and will fail – that is, the result cannot have more digits than the value of precision in effect for the operation, and will not be rounded."

http://speleotrove.com/decimal/daops.html#refdivint

---

*[context.go, line 462 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNOFKJNK5kzXEfo8c5:-Kfh1zNB2f1S4i7i3xhr:bxcxqab) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L462)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Don't we need to invert the sign when y is negative? IIRC, floating point division should satisfy -x / y = x / -y = -(x / y).
</blockquote></details>

The spec says: "The sign of the result, if non-zero, is the same as that of the original dividend."

http://speleotrove.com/decimal/daops.html#refremain

So I guess we don't satisfy that.

---

*[context.go, line 617 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQ1pXL8lG8m_oeko3:-Kfh2CCSUm991SGNUBF_:b-f0tukj) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L617)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Does this `Cbrt` return NaN on negative inputs? Should it?
</blockquote></details>

It does (see the changes to cuberoot-apd.decTest), and it should do exactly the same thing Sqrt does for these inputs, I think.

---

*[context.go, line 891 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQOAkhPNeqWCb5Lkf:-Kfh44sByeiBoSAMuBrR:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L891)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

`logSpecials`?
</blockquote></details>

Done.

---

*[context.go, line 934 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNR3jlRcI5Gz3s9MMf:-Kfh4FBOCmqfgoGYBljM:b-nxrh99) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L934)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

How many of these functions need to set the Form field? I guess the transcendental functions don't because they're computed by iterative algorithms that use lower-level functions.
</blockquote></details>

All functions must set the Form field. I added stuff to gda_test.go that fills in a bogus Form value to make sure it always gets set. And yes, the transcendental functions don't need to explicitly set it because of the simpler functions that do. But overall I'm convinced that all branches set Form because of the gda_test modification described.

---

*[context.go, line 1214 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg78nhYXK447Gk40Uq:-Kfh5DGbetzlihPawjhu:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L1214)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Could this check go in `Quantize`? I'm not sure why we would test `d.NumDigits()` when `d` is not a number.
</blockquote></details>

Done.

---

*[decimal.go, line 43 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg7xk9PojutApH5n2Y:-Kfh5jLZwhM9QjrQO8RF:bnzgi8r) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L43)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

I had to read this comment a couple times to get it. Perhaps "`CmpTotal` assumes that the order of these constants reflects the total order on decimals"? `cmpOrder` seems to make further assumptions, which may be worth mentioning.
</blockquote></details>

Done. But I'm not sure what further assumptions you are referring to.

---

*[decimal.go, line 114 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9ByW6akrodhhWEE6:-Kfh7V7eUyqvbaZC1970:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L114)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Rather than count here, perhaps introduce a function like

```
func ConsumePrefix(s, prefix string) (string, bool) {
  if strings.HasPrefix(s, prefix) {
    return s[len(prefix):], true
  }
  return s, false
}
```

and use throughout.
</blockquote></details>

Done.

---

*[decimal.go, line 132 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9lghL5ky0eai33el:-Kfh6a_LSXw6lJbTbF1p:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L132)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Couldn't this be done earlier?
</blockquote></details>

Done.

---

*[decimal.go, line 136 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg9yS5TfQ9qMEbjwRh:-Kfh6qjK4u7U_4K56poK:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L136)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

We're only looking for lowercase e now, so we can use `IndexByte`.
</blockquote></details>

Done.

---

*[decimal.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgAPHyXrFdam38VzGo:-Kfh8MHZvWT9myFE2CM1:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L244)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Is it worth pulling out the duplicate handling for specials and negative numbers between this function and the previous function?
</blockquote></details>

Done.

---

*[decimal.go, line 511 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgCFXYKSHFaqrmtXAi:-Kfh8gq1rCqYxcYa2R9o:b9f1k8i) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L511)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Unless I'm mistaken, `cmpOrder` returns `0` for all finite decimals, both negative and positive, so `d` and `x` don't necessarily have the same sign. I guess this works anyway because `Cmp` handles the signs too. See below.
</blockquote></details>

Good catch. I didn't put together that `Finite == -Finite` since it's 0.

---

*[decimal.go, line 536 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgCRLK6dcFFEOzuMuz:-KfhLMBgyrBe6ITsuoaa:bib8zk4) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L536)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Probably `v = ^v` (or `v = -(v + 1)` if that's too clever); see above.
</blockquote></details>

This triggered a huge exploration that uncovered various bugs in `CmpTotal` and `add`, including the elusive RoundFloor negative thing that wasn't tested. Also another way to achieve this that I find a bit easier to understand, is doing `v := int(d.Form) + 1` above. The `+1` was more intuitive for me instead of the `^`. Good find.

---

*[decimal.go, line 568 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgF3-W9x8cgS66XbPm:-Kfj5xpZncEvaa34rzK4:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L568)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Wait to define these until after checking the signs and leave a comment explaining them. It looks strange at first that the sign tests use ±1 while the rest uses the "constants".
</blockquote></details>

Done.

---

*[decimal.go, line 620 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgFOLAvhU8jxyBfgJj:-Kfj7uEGCfs3Y497sQar:bilm3ti) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L620)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

I'm confused by the inner functions. We never seem to set `v` and then not return true, so couldn't we just return from the outer function instead?
</blockquote></details>

I think I did this for a better reason that then got refactored away. Removed this functions.

---

*[decimal.go, line 666 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgFsONXJz4UINSDtWy:-Kfj87n-AL4qJ_C9N57u:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L666)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

I prefer

```
if d.Form == Finite && d.Coeff.Sign() == 0 {
  return 0
}
if d.Negative {
  return -1
}
return 1
```
</blockquote></details>

Done.

---

*[decimal.go, line 686 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgG8VuUyGlgP60Sq0r:-Kfj8Gd44v0dtwJbWKsh:bvy7pa6) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L686)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Is the Form check necessary?
</blockquote></details>

I thought it was, but since Sign() can only return 0 if d is Finite, I agree it isn't necessary. Removed.

---

*[gda_test.go, line 449 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgHLQ_Vd7EHlnzxI0V:-Kfj9aK-LIpPcyz7Pz3X:bb0skd7) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/gda_test.go#L449)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Doesn't `Quantize` handle infinite operands now?
</blockquote></details>

This is checking the second operand. Quantize takes an int32 instead of a Decimal as the second operand, and so we skip all tests where the second operand isn't finite.

---

*[gda_test.go, line 627 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfgHjqkQQIrLQt6NN5k:-Kfj9q_JT2AvRQWF1uUo:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/gda_test.go#L627)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

`strings.EqualFold`? Kind of pedantic here admittedly.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/33#-:-KfgsG6YymFF2E8DplSR:b-i05c6h)*

eisenstatdavid:
:lgtm:

Thanks for writing additional tests.

---

Reviewed 1 of 8 files at r1, 1 of 1 files at r4, 7 of 7 files at r5.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[context.go, line 125 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfMnKLlWi9wjmkKyeaq:-KflJIyNwCku-kbFMJFp:blca0d0) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L125)):*

I spent a few minutes trying to get this to work (that is, all of your comments about this function) but failed, as some small subset of tests always failed. I agree it seems like there's a simpler solution than what we have here but I currently think it should be done in another PR. I've created #37 to track this.
</blockquote></details>

OK, I'll try again out of band.

---

*[context.go, line 280 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfN90DxAJrLbSd27tWu:-KflJoDQMQO45rLBRJxG:b-wgomvk) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L280)):*

I don't think we need a `y == 0` check here because it's covered in the tests. For example (https://github.com/cockroachdb/apd/blob/c7ade124510c862bab38aa30f412aebe6c7c2abd/testdata/divide.decTest#L597):

```
divx784 divide  Inf   0     ->  Infinity
```

(This is `Inf / 0 == Inf`)

The precision check was added by me to prevent footgun problems. I think that this block should stay above the precision check because in these cases we can produce correct, fast results.
</blockquote></details>

I guess this is a special case of positive decimal divided by zero.

---

*[context.go, line 617 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQ1pXL8lG8m_oeko3:-KflL6okSOSm2lGpw-Lo:b8dbs17) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L617)):*

It does (see the changes to cuberoot-apd.decTest), and it should do exactly the same thing Sqrt does for these inputs, I think.
</blockquote></details>

Weird - - I would have expected it to copy the sign of its input, since x^3 - a has exactly one real root for all a. I don't remember if SQL has cbrt, but if it does, we should think about the alignment between this function and IEEE cbrt, which accepts negative numbers.

---

*[context.go, line 1257 at r5](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KflMXp01fFmWEfV_iUl:-KflMDO7rsN-mEMj8G_N:b5xc2oh) ([raw file](https://github.com/cockroachdb/apd/blob/3763b98152864d363a259995d0b4c0b9cded6b0b/context.go#L1257)):*
> ```Go
> 	}
> 	d.Exponent = exp
> 	d.Negative = neg
> ```

Do we mutate either v or neg above? Wondering why this isn't `d.Negative = v.Negative` or suppressed entirely by dint of the `d.Set(v)` above.

I guess I can answer my own question: because d and v may be aliased.

---

*[decimal.go, line 43 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-Kfg7xk9PojutApH5n2Y:-KflMhs6ZdUEgtVKXIj8:bcyhkha) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/decimal.go#L43)):*

Done. But I'm not sure what further assumptions you are referring to.
</blockquote></details>

That for all constant values c1 and c2, -(c1 + 1) is less than c2 + 1. Maybe not comment-worthy since I don't expect this enumeration to change, but a little unusual.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/33#-:-KflNexocuyWq7uJ1JJa:b-bt5z5n)*

mjibson:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[context.go, line 617 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KfNQ1pXL8lG8m_oeko3:-KflrHcPaqQ_XYy7ovHt:bl1bbjx) ([raw file](https://github.com/cockroachdb/apd/blob/9e76a70e3e13c4de1f89c0f54814851bbae3b93b/context.go#L617)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Weird - - I would have expected it to copy the sign of its input, since x^3 - a has exactly one real root for all a. I don't remember if SQL has cbrt, but if it does, we should think about the alignment between this function and IEEE cbrt, which accepts negative numbers.
</blockquote></details>

You are correct. Cbrt should support negative numbers. I didn't think that through completely. I've opened #38 to track this.

---

*[context.go, line 1257 at r5](https://reviewable.io:443/reviews/cockroachdb/apd/33#-KflMXp01fFmWEfV_iUl:-KflsE_S5mPRev43YGv7:b-20p7fn) ([raw file](https://github.com/cockroachdb/apd/blob/3763b98152864d363a259995d0b4c0b9cded6b0b/context.go#L1257)):*
<details><summary><i>Previously, eisenstatdavid (David Eisenstat) wrote…</i></summary><blockquote>

Do we mutate either v or neg above? Wondering why this isn't `d.Negative = v.Negative` or suppressed entirely by dint of the `d.Set(v)` above.

I guess I can answer my own question: because d and v may be aliased.
</blockquote></details>

I was able to remove this completely. The problem was the `d.SetCoefficient(0)` in the !d.IsZero block. Using Coeff.SetInt64 instead doesn't squash the negative setting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/33)*

PR #35, opened 2017-03-17
DO NOT MERGE sqrt: add review comments.

arjunravinarayan:
This is a dummy PR to simulate a code review on the square root code
in the existing APD library.

Apologies that this is so late.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/35)

### DIFF
--- a/context.go
+++ b/context.go
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
 }
 
 // Sqrt sets d to the square root of x.
+// Sqrt uses the Babylonian method for computing the square root, which uses
+// O(log p) steps for p digits of precision. We follow the algorithm in
+// Properly Rounded Variable Precision Square Root by T. E. Hull and A. Abrham,
+// ACM Transactions on Mathematical Software, Vol 11 #3, pp229–237, ACM, September 1985.
 func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
-	// See: Properly Rounded Variable Precision Square Root by T. E. Hull
-	// and A. Abrham, ACM Transactions on Mathematical Software, Vol 11 #3,
-	// pp229–237, ACM, September 1985.
 
 	switch x.Sign() {
 	case -1:
-		res := InvalidOperation
-		return c.goError(res)
+		return c.goError(InvalidOperation)
 	case 0:
 		d.SetCoefficient(0)
 		d.Exponent = 0
 		return 0, nil
 	}
+	// x > 0.
 
-	// Use same precision as in decNumber.
-	workp := c.Precision + 1
-	if nd := uint32(x.NumDigits()); workp < nd {
-		workp = nd
-	}
-	if workp < 7 {
-		workp = 7
+	// TODO(mjibson): Also it would be good if we could keep the number of
+	// variables in scope low. This refactoring doesn't quite work, but perhaps
+	// you can refactor the calls below to use nc.Precision instead of keeping
+	// workp in scope throughout.
+	// TODO(mjibson): Also, what's nc?
+	var nc *Context
+	{
+		// workp is the number of digits of precision used.
+		// We use same precision as in decNumber.
+		workp := c.Precision + 1
+		if nd := uint32(x.NumDigits()); workp < nd {
+			workp = nd
+		}
+		if workp < 7 {
+			workp = 7
+		}
+		nc = c.WithPrecision(workp)
 	}
 
-	f := new(Decimal).Set(x)
-	nd := x.NumDigits()
-	e := nd + int64(x.Exponent)
-	f.Exponent = int32(-nd)
+	// TODO(mjibson): Too many variables without comments/definitions. And nd is a temporary variable that goes out of scope, perhaps it can be closed off? I've given it a try but I'm not too happy with this either... Also, these are my *guesses* for what these variables represent.
+	// f is ...
+	var f *Decimal
+	// e is the number of digits in x.
+	var e int64
+	// ed is ...
+	var ed ErrDecimal
+	{
+		f = new(Decimal).Set(x)
+		nd := x.NumDigits()
+		e = nd + int64(x.Exponent)
+		f.Exponent = int32(-nd)
+		nc.Rounding = RoundHalfEven
+		ed = MakeErrDecimal(nc)
+	}
+
+	// TODO(mjibson): This is my best guess for what's going on, but I'm unsure, so please check:
+
+	// approx stores our first approximation for sqrt(x). We have two guesses: if e is even, then we use .259 + 0.819*f. Oterwise we use 0.0819 + 2.59*f.
+	// TODO(mjibson): Why? I think it's because these are the geometric midpoint "best guesses" for 1 <= f < 10 and 10 <= f < 100 respectively, but I'm not sure.
 	approx := new(Decimal)
-	nc := c.WithPrecision(workp)
-	nc.Rounding = RoundHalfEven
-	ed := MakeErrDecimal(nc)
 	if e%2 == 0 {
 		approx.SetCoefficient(819).SetExponent(-3)
 		ed.Mul(approx, approx, f)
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -380,13 +404,16 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
...
+	// quadratically, which we keep track of in p.
 	p := uint32(3)
 	tmp := new(Decimal)
 
 	// The algorithm in the paper says to use c.Precision + 2. decNumber uses
 	// workp + 2. But we use workp + 5 to make the tests pass. This means it is
 	// possible there are inputs we don't compute correctly and could be 1ulp off.
-	for maxp := workp + 5; p != maxp; {
+	for maxp := nc.Precision + 5; p != maxp; {
+		// TODO(mjibson): why the minus 2?
```
mjibson: Because that's what's written in the paper. Did I miss something?

context.go
```diff
@@ -921,6 +949,8 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 
 	// decNumber sets the precision to be max(x digits + exponent, c.Precision)
 	// + 4. 6 is used as the exponent digits.
+	// TODO(mjibson): what is decNumber, and what is the source of the +4?
```
mjibson: decNumber is the reference implementation from the GDA site. decNumber is referred to in a few parts and I don't think it needs a specific explanation here. I'm not sure where they got the + 4. The comment about the 6 is the same as in the decNumber code, and I'm not able to explain that either. \o/ But the tests pass so...?

context.go
```diff
@@ -967,6 +997,9 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 
 // Quantize adjusts and rounds v as necessary so it is represented with
 // exponent exp and stores the result in d.
+// TODO(mjibson): "adjusts"?
```
mjibson: If you have a specific suggestion I welcome it. This function is pretty hard to describe. There are runnable examples in the docs, though, that make it easy to understand.

context.go
```diff
@@ -380,13 +404,16 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
 		ed.Add(approx, approx, New(819, -4))
 	}
 
+	// TODO(mjibson): Now we repeatedly improve approx. Our precision improves
```
mjibson: I'm not convinced it's best to add comments like this, that are describing how or why the paper chose this algorithm. I think those descriptions are best left in the paper. I think comments in our implementation should focus on differences between us and the paper, or how we do things in Go that look slightly different than the paper.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
 	}
-	if workp < 7 {
-		workp = 7
+
+	// TODO(mjibson): Too many variables without comments/definitions. And nd is a temporary variable that goes out of scope, perhaps it can be closed off? I've given it a try but I'm not too happy with this either... Also, these are my *guesses* for what these variables represent.
+	// f is ...
+	var f *Decimal
+	// e is the number of digits in x.
+	var e int64
+	// ed is ...
```
mjibson: `ed` is used consistently throughout as the error accumulator for internal operations. I think a comment describing that would be more annoying than useful since it's the same everywhere.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
+		if workp < 7 {
+			workp = 7
+		}
+		nc = c.WithPrecision(workp)
 	}
-	if workp < 7 {
-		workp = 7
+
+	// TODO(mjibson): Too many variables without comments/definitions. And nd is a temporary variable that goes out of scope, perhaps it can be closed off? I've given it a try but I'm not too happy with this either... Also, these are my *guesses* for what these variables represent.
+	// f is ...
```
mjibson: Both `f` and `e` are the same as in the paper. I don't think it makes sense to describe them here.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
+		}
+		if workp < 7 {
+			workp = 7
+		}
+		nc = c.WithPrecision(workp)
 	}
-	if workp < 7 {
-		workp = 7
+
+	// TODO(mjibson): Too many variables without comments/definitions. And nd is a temporary variable that goes out of scope, perhaps it can be closed off? I've given it a try but I'm not too happy with this either... Also, these are my *guesses* for what these variables represent.
```
mjibson: `nd` is the only variable I can see as benefiting from this, but it doesn't even gain that much because it requires a `var e blah` above it, which I don't think is better. All the others are used throughout this function because that's how they work in the paper or decNumber.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
-	// Use same precision as in decNumber.
-	workp := c.Precision + 1
-	if nd := uint32(x.NumDigits()); workp < nd {
-		workp = nd
+	// x > 0.
+
+	// TODO(mjibson): Also it would be good if we could keep the number of
+	// variables in scope low. This refactoring doesn't quite work, but perhaps
+	// you can refactor the calls below to use nc.Precision instead of keeping
+	// workp in scope throughout.
```
mjibson: I'm not sure how to do this without an extra variable because nc.Precision needs to be set to workp later down in the function, so we have to store it somewhere.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
 }
 
 // Sqrt sets d to the square root of x.
+// Sqrt uses the Babylonian method for computing the square root, which uses
```
mjibson: I really like this basic description of the algorithm and the big O for the number of steps. However I'm going to leave the reference as is to be consistent with the other references, none of which show up in the godoc. If you think it should be in the godoc, we can have that discussion and move them all at the same time.

arjunravinarayan: I'm okay with that. I didn't realize the change in position would have godoc effects.

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
 func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
-	// See: Properly Rounded Variable Precision Square Root by T. E. Hull
-	// and A. Abrham, ACM Transactions on Mathematical Software, Vol 11 #3,
-	// pp229–237, ACM, September 1985.
 
 	switch x.Sign() {
 	case -1:
-		res := InvalidOperation
-		return c.goError(res)
+		return c.goError(InvalidOperation)
```
arjunravinarayan: OK.

context.go
arjunravinarayan: OK

context.go
```diff
@@ -336,38 +336,62 @@ func (c *Context) Rem(d, x, y *Decimal) (Condition, error) {
...
-	workp := c.Precision + 1
-	if nd := uint32(x.NumDigits()); workp < nd {
-		workp = nd
+	// x > 0.
+
+	// TODO(mjibson): Also it would be good if we could keep the number of
+	// variables in scope low. This refactoring doesn't quite work, but perhaps
+	// you can refactor the calls below to use nc.Precision instead of keeping
+	// workp in scope throughout.
+	// TODO(mjibson): Also, what's nc?
```
arjunravinarayan: OK.

context.go
arjunravinarayan: OK.

context.go
arjunravinarayan: OK. Nevertheless, I think the paper is too terse! Not blaming you, but I couldn't make sense of what those variables were from both the code and the paper. Still, you can leave it alone. We can't be rewriting papers as well.

context.go
arjunravinarayan: OK.

context.go
arjunravinarayan: OK, that's a stylistic choice that I can live with, although I would do it differently.

context.go
arjunravinarayan: No, I missed it.

context.go
```diff
@@ -399,7 +426,8 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
 		// approx = 0.5 * (approx + f / approx)
 		ed.Mul(approx, tmp, decimalHalf)
 	}
-	nc.Precision = workp + 1
+	// TODO(mjibson): why the plus 1?
```
arjunravinarayan: :+1:!

context.go
arjunravinarayan: OK on decNumber.
The 4 and 6 I'm less OK with, perhaps we should leave a todo here making it clear that these are cargo-cult magic constants (since the other magic constants are clearly documented).

context.go
arjunravinarayan: Discretize?

context.go
```diff
@@ -967,6 +997,9 @@ func (c *Context) Pow(d, x, y *Decimal) (Condition, error) {
 
 // Quantize adjusts and rounds v as necessary so it is represented with
 // exponent exp and stores the result in d.
+// TODO(mjibson): "adjusts"?
+// TODO(mjibson): Could use a comment (not just here), or a wrapper Function
+// since you call this sometimes with the same argument for both d and v and that scares me.
```
arjunravinarayan: I'm not at all concerned about correctness because the testing throughout is ridiculously good, but it is weird to see, but that's fine. I'm OK with it.

arjunravinarayan:
Looks like everything is addressed, and we can continue the conversation at #36.

PR #43, opened 2017-03-21
Remove unused block in sqrt

mjibson:
Neither of the branches that modified approx here were taken in any
tests, so the tests still pass without them. Even though this is
described in the paper, I think it is safe to remove.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/43)

### DIFF
--- a/context.go
+++ b/context.go
@@ -531,25 +531,6 @@ func (c *Context) Sqrt(d, x *Decimal) (Condition, error) {
 		// approx = 0.5 * (approx + f / approx)
 		ed.Mul(approx, tmp, decimalHalf)
 	}
-	nc.Precision = workp
-	dp := int32(c.Precision)
-	approxsubhalf := new(Decimal)
-	ed.Sub(approxsubhalf, approx, New(5, -1-dp))
-	nc.Rounding = RoundUp
-	ed.Mul(approxsubhalf, approxsubhalf, approxsubhalf)
-	if approxsubhalf.Cmp(f) > 0 {
-		// TODO(mjibson): this branch is never taken in tests, why? Can it be removed?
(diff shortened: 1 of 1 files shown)


arjunravinarayan:
I'm somewhat less happy with this given that your philosophy is "we don't document algorithms because we cite and then follow the paper". Can you add a comment about why you think this final bit of the algorithm, while listed in the paper, is not necessary?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/43#-:-Kfwv39xBqQZm0GXLiyj:b-4j8wg6)*

mjibson:
I agree. I've added a comment describing this and the justification. I'm not 100% certain that my reasoning is correct, but it's the best I've come up with. If you have more insight about why these branches were never taken I'd like to use that.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/43#-:-KfwvuLvSCQtxSNYPmcY:b-dznere)*

mjibson:
ping @arjunravinarayan

arjunravinarayan:
:lgtm:

Comment is good enough for me, I don't see any better justification.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/43#-:-KgG-bjQ64yUs1F3gZBu:b-4c7y8g)*

mjibson:
TFTR.

Issue #44, opened 2017-03-23
Implement Sqrt(f *big.Float) *big.Float


plutov:
Standard `math/big` package has no method to calculate `Sqrt` of big sort. I have implemented this function for this exercise - https://github.com/plutov/practice-go/tree/master/sumdecimal.

I can contribute it into `apd` if it's necessary via PR.

mjibson:
See comment in #45

PR #45, opened 2017-03-23
#44. Sqrt function

plutov:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/apd/45)

### DIFF
--- a/bigfloat.go
+++ b/bigfloat.go
@@ -1,0 +1,127 @@
+package apd
+
+import (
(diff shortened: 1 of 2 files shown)


mjibson:
I like the direction of this: to fill out some of the methods not present in big.Float. However apd is not the appropriate project to put this stuff in. I recommend starting another project or finding an existing one to add to.

PR #55, opened 2017-06-13
some Cmp optimizations

RaduBerinde:
Together these improve a big sorting query (`SELECT * FROM tpch.orders ORDER BY o_totalprice`) by about 7 seconds (35s to 28s) 

### skip border comparison in NumDigits when possible
Avoid the comparison with the border value when there is only one possible value
for a given bit length (~2 out of 3 times).

### optimize comparison
If two numbers have the same exponent, just compare the coefficients.
Otherwise, don't allocate two big Ints when we only need to scale one of them.
And put the big.Int slice on the stack, not on the heap (depends on go's escape
analysis if this actually stays on the stack though).

### DIFF
--- a/table.go
+++ b/table.go
@@ -25,12 +25,9 @@ import "math/big"
 // Using this proof, for a given digit count, the map will return the lower number
 // of decimal digits (k) the binary digit count could represent, along with the
 // value of the border between the two decimal digit counts (10^k).
-const (
-	digitsTableSize = 128
-	digitsTableLen  = digitsTableSize + 1
-)
+const digitsTableSize = 128
 
-var digitsLookupTable [digitsTableLen]tableVal
+var digitsLookupTable [digitsTableSize + 1]tableVal
 
 type tableVal struct {
 	digits  int64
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -70,6 +70,15 @@ func NumDigits(b *big.Int) int64 {
 
 	if bl < digitsTableLen {
 		val := digitsLookupTable[bl]
+		// In general, we either have val.digits or val.digits+1 digits and we have
+		// to compare with the border value. But that's not true for all values of
+		// bl: in particular, if bl+1 maps to the same number of digits, then we
+		// know for sure we have val.digits and we can skip the comparison.
+		// This is the case for about 2 out of 3 values.
+		if bl+1 < digitsTableLen && digitsLookupTable[bl+1].digits == val.digits {
```
mjibson: You could use `bl < digitsTableSize` to avoid the addition.

RaduBerinde: Done. Removed `digitsTableLen` altogether, it's strange to have both of these constants.

mjibson:
For the archives:

```
name                 old time/op    new time/op    delta
GDA/compare-12          475µs ± 8%     433µs ± 6%   -8.84%  (p=0.002 n=10+9)
GDA/comparetotal-12     473µs ±11%     443µs ± 6%   -6.45%  (p=0.005 n=10+10)

name                 old alloc/op   new alloc/op   delta
GDA/compare-12         19.5kB ± 0%     6.0kB ± 0%  -69.21%  (p=0.000 n=10+9)
GDA/comparetotal-12    19.8kB ± 0%     6.1kB ± 0%  -69.17%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
GDA/compare-12            406 ± 0%       125 ± 0%  -69.21%  (p=0.000 n=10+10)
GDA/comparetotal-12       412 ± 0%       127 ± 0%  -69.17%  (p=0.000 n=10+10)
```

PR #59, opened 2017-06-30
Improvements

mjibson:
Various improvements. See http://speleotrove.com/decimal/daops.html#refrtoix for the naming choice on the ToIntegral functions.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -680,37 +680,65 @@ func (d *Decimal) IsZero() bool {
 }
 
 // Modf sets integ to the integral part of d and frac to the fractional part
-// such that d = integ+frac. If d is negative, both integ or frac will be
-// either 0 or negative. integ.Exponent will be >= 0; frac.Exponent will be
-// <= 0.
+// such that d = integ+frac. If d is negative, both integ or frac will be either
+// 0 or negative. integ.Exponent will be >= 0; frac.Exponent will be <= 0.
+// Either argument can be nil, preventing it from being set.
 func (d *Decimal) Modf(integ, frac *Decimal) {
+	if integ == nil && frac == nil {
+		return
+	}
+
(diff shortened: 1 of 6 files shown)


decimal.go
```diff
@@ -677,37 +677,61 @@ func (d *Decimal) IsZero() bool {
...
-	frac.Negative = neg
-	integ.Negative = neg
+
+	var icoeff *big.Int
+	if integ != nil {
+		icoeff = &integ.Coeff
+		integ.Exponent = 0
+		integ.Negative = neg
+	} else {
+		icoeff = new(big.Int)
```
knz: Skip the allocation if frac == nil

decimal.go
```diff
@@ -677,37 +677,61 @@ func (d *Decimal) IsZero() bool {
...
+	}
+
+	if frac != nil {
+		icoeff.QuoRem(&d.Coeff, e, &frac.Coeff)
+		frac.Exponent = d.Exponent
+		frac.Negative = neg
+	} else {
+		icoeff.Quo(&d.Coeff, e)
```
knz: Skip the computation if integ == nil

decimal.go
mjibson: This is the `integ == nil` branch, and we already checked if both integ and frac were nil above, so frac can never be nil in this branch.

decimal.go
mjibson: Same as above: we are in frac == nil, which means integ must not be nil since they both can't be.

decimal.go
knz: Sorry I hadn't seen the check at the beginning. Please add the explanatory comment here to repeat the assumption.

mjibson: Done.

knz:
thanks!

PR #60, opened 2017-07-01
add Format method

### LINKED ISSUE #19: add a FormatStandard(buf *bytes.Buffer) method

mjibson:
Supports e, E, f, g, G, s, v.

ToSci has been removed and is now just String, which wraps .Text('G').
ToStandard has been removed, but can still be produced using the 'f'
format. The precision field is not supported because that could require
rounding and we don't have a context. Instead, the format methods here
always display the full precision. Although big.Float doesn't implement
's', I think we safely can because the precision is always fully
displayed.

Fixes #19


mjibson:
Ping @knz.

decimal_test.go
```diff
@@ -388,28 +388,154 @@ func TestFloor(t *testing.T) {
...
+			e: "-0e-1",
+			E: "-0E-1",
+		},
+		"0E+2": {
+			e: "0e+2",
+			f: "000",
+			g: "0e+2",
+		},
+	}
+	verbs := []string{"e", "E", "f", "g", "G"}
```
knz: to make linters happy, please consider using `verbs := []string{"%e", "%E", ...}` here and `Sprintf(v, d)` below.

format.go
```diff
@@ -0,0 +1,204 @@
...
+	case 'f':
+		return fmtF(buf, d, digits)
+	case 'g', 'G':
+		// See: http://speleotrove.com/decimal/daconvs.html#reftostr
+		const adjExponentLimit = -6
+		adj := int(d.Exponent) + (len(digits) - 1)
+		if d.Exponent <= 0 && adj >= adjExponentLimit {
+			return fmtF(buf, d, digits)
+		}
+		return fmtE(buf, fmt+'e'-'g', d, digits)
```
knz: I am not familiar with this syntax in Go `fmt + 'e' - 'g'` what does it mean? In C that would be equivalent to `fmt - 2`

format.go
```diff
@@ -0,0 +1,204 @@
...
+	case 'e', 'E', 'f', 'g', 'G':
+		// nothing to do
+	case 'F':
+		// (*Decimal).Text doesn't support 'F'; handle like 'f'
+		format = 'f'
+	case 'v', 's':
+		// handle like 'G'
+		format = 'G'
+	default:
+		fmt.Fprintf(s, "%%!%c(*apd.Decimal=%s)", format, d.String())
```
knz: Add a test case for this format too.

decimal_test.go
mjibson: done

format.go
mjibson: This is copied from https://github.com/golang/go/blob/fcc35147d5940d5b077586c02a97894c3edf870d/src/math/big/ftoa.go?utf8=%E2%9C%93#L148

It needs to convert the either `g` or `G` into a `e` or `E` since that's what `fmtE` expects. So it is indeed `fmt - 2`, but attempting to do that in a way that illustrates the intention.

format.go
mjibson: done

format.go
knz: Thank you! Perhaps a comment would help too :)

knz:
:+1: great functionality!

PR #61, opened 2017-07-07
use correct cbrt estimation const

mjibson:
C3 was incorrectly using the C2 constant. Luckily this was only used
for the initial approximation, and thus affected only performance,
not correctness.


mjibson:
(As noted in the code, see https://people.freebsd.org/~lstewart/references/apple_tr_kt32_cuberoot.pdf for the consts.)

RaduBerinde:
LGTM

PR #64, opened 2017-07-12
add NullDecimal

### LINKED ISSUE #63: Consider implementing NullDecimal for use with database/sql

mjibson:
Fixes #63

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -772,3 +772,38 @@ func (d *Decimal) Scan(src interface{}) error {
 		return errors.Errorf("could not convert %T to Decimal", src)
 	}
 }
+
+// NullDecimal represents a string that may be null. NullDecimal implements
+// the database/sql.Scanner interface so it can be used as a scan destination:
+//
+//  var d NullDecimal
+//  err := db.QueryRow("SELECT num FROM foo WHERE id=?", id).Scan(&d)
+//  ...
+//  if d.Valid {
+//     // use d.Decimal
+//  } else {
+//     // NULL value
+//  }
+//
+type NullDecimal struct {
+	Decimal Decimal
+	Valid   bool // Valid is true if Decimal is not NULL
+}
+
+// Scan implements the database/sql.Scanner interface.
+func (nd *NullDecimal) Scan(value interface{}) error {
+	if value == nil {
+		nd.Valid = false
+		return nil
+	}
+	nd.Valid = true
+	return nd.Decimal.Scan(value)
+}
+
+// Value implements the database/sql/driver.Valuer interface.
+func (nd NullDecimal) Value() (driver.Value, error) {
+	if !nd.Valid {
+		return nil, nil
+	}
+	return nd.Decimal.Value()
+}
--- a/sql_test.go
+++ b/sql_test.go
@@ -38,7 +38,10 @@ func TestSQL(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if _, err := db.Exec("create table if not exists d (v decimal); delete from d"); err != nil {
+	if _, err := db.Exec("drop table if exists d"); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := db.Exec("create table d (v decimal)"); err != nil {
 		t.Fatal(err)
 	}
 	if _, err := db.Exec("insert into d values ($1)", a); err != nil {
@@ -48,16 +51,27 @@ func TestSQL(t *testing.T) {
 		t.Fatal(err)
 	}
 	var b, c, d Decimal
-	if err := db.QueryRow("select v, v::text, v::int, v::float from d").Scan(a, &b, &c, &d); err != nil {
+	var nd NullDecimal
+	if err := db.QueryRow("select v, v::text, v::int, v::float, v from d").Scan(a, &b, &c, &d, &nd); err != nil {
 		t.Fatal(err)
 	}
 	want, _, err := NewFromString("123556700")
 	if err != nil {
 		t.Fatal(err)
 	}
-	for i, v := range []*Decimal{a, &b, &c, &d} {
+	for i, v := range []*Decimal{a, &b, &c, &d, &nd.Decimal} {
 		if v.Cmp(want) != 0 {
-			t.Fatalf("%d: unexpected: %s, want: %s", i, v.ToStandard(), want.ToStandard())
+			t.Fatalf("%d: unexpected: %s, want: %s", i, v.String(), want.String())
 		}
+	}
+
+	if _, err := db.Exec("update d set v = NULL"); err != nil {
+		t.Fatal(err)
+	}
+	if err := db.QueryRow("select v from d").Scan(&nd); err != nil {
+		t.Fatal(err)
+	}
+	if nd.Valid {
+		t.Fatal("expected null")
 	}
 }


knz:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[decimal.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kot5AlYHGnGVREdzKxZ:-Kot5AlYHGnGVREdzKx_:bmcwxrv) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L800)):*
> ```Go
> 	}
> 	nd.Valid = true
> 	return nd.Decimal.Scan(value)
> ```

`Valid` should only be set if `Scan` returns no error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64)*

mjibson:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[decimal.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kot5AlYHGnGVREdzKxZ:-KouqAW1wfxHTmDT7h6F:b-sxmm0f) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L800)):*
<details><summary><i>Previously, knz (kena) wrote…</i></summary><blockquote>

`Valid` should only be set if `Scan` returns no error.
</blockquote></details>

This matches NullString in the stdlib, so I think it's correct: https://golang.org/src/database/sql/sql.go?s=4428:4479#L157

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64)*

mjibson:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[decimal.go, line 789 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kov2L2ebNyktamz-YcY:-Kov2L2ebNyktamz-YcZ:btnovmt) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L789)):*
> ```Go
> //
> type NullDecimal struct {
> 	Decimal Decimal
> ```

Also what do you think about this? Should it be `*Decimal` instead? That forces Scan to allocate. I'm not sure which is better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64)*

knz:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[decimal.go, line 789 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kov2L2ebNyktamz-YcY:-KovkMj04JFbcF15GsEg:brhh0qv) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L789)):*

Also what do you think about this? Should it be `*Decimal` instead? That forces Scan to allocate. I'm not sure which is better.
</blockquote></details>

So I think it depends on what the caller will do with the `NullDecimal`. If they merely check `Valid` and then expect to "carry the `Decimal` field away" somewhere else then `*Decimal` might be better.

---

*[decimal.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kot5AlYHGnGVREdzKxZ:-KovkaAEqSx8OWY4tBUx:b23u0r) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L800)):*

This matches NullString in the stdlib, so I think it's correct: https://golang.org/src/database/sql/sql.go?s=4428:4479#L157
</blockquote></details>

Okay then it must be right, but I do not fully understand why. It seemed to me that in Go we usually don't modify the destination variable in case of error. Maybe the go sql package does things differently.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64)*

mjibson:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[decimal.go, line 789 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kov2L2ebNyktamz-YcY:-Kowji_hnu9fqGprcsWf:bv2wy2m) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L789)):*
<details><summary><i>Previously, knz (kena) wrote…</i></summary><blockquote>

So I think it depends on what the caller will do with the `NullDecimal`. If they merely check `Valid` and then expect to "carry the `Decimal` field away" somewhere else then `*Decimal` might be better.
</blockquote></details>

I tried out a full implementation with *Decimal and it has an explicit

```
	if nd.Decimal == nil {
		nd.Decimal = new(Decimal)
	}
```

in Scan which I just really don't like. It feels like the allocation requirement of that is not good. I think leaving it as is is good for now.

---

*[decimal.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kot5AlYHGnGVREdzKxZ:-KowgqXbyIIJrNAPS4Lv:b-a6wxtu) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L800)):*
<details><summary><i>Previously, knz (kena) wrote…</i></summary><blockquote>

Okay then it must be right, but I do not fully understand why. It seemed to me that in Go we usually don't modify the destination variable in case of error. Maybe the go sql package does things differently.
</blockquote></details>

I think the standard is that if an error is returned, the other variable is not in a known state (unless specified by the docs).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64)*

knz:
:lgtm:

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[decimal.go, line 789 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kov2L2ebNyktamz-YcY:-Kp0eqqbVDCP7DVoduJX:b-97hr5d) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L789)):*

I tried out a full implementation with *Decimal and it has an explicit

```
	if nd.Decimal == nil {
		nd.Decimal = new(Decimal)
	}
```

in Scan which I just really don't like. It feels like the allocation requirement of that is not good. I think leaving it as is is good for now.
</blockquote></details>

Yes, I agree.

---

*[decimal.go, line 800 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/64#-Kot5AlYHGnGVREdzKxZ:-Kp0esrIhUgZtzkELAJO:b-jhkb7z) ([raw file](https://github.com/cockroachdb/apd/blob/850c06e03ba07d93999d843eb9f8e83389c03f36/decimal.go#L800)):*

I think the standard is that if an error is returned, the other variable is not in a known state (unless specified by the docs).
</blockquote></details>

Aye. TIL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/64#-:-Kp0eumyQhetdGFsSOUT:bnfp4nl)*

PR #67, opened 2017-08-25
Support marshaling/unmarshaling JSON

dhui:


decimal.go
```diff
@@ -773,6 +774,24 @@ func (d *Decimal) Scan(src interface{}) error {
 	}
 }
 
+var jsonNull = []byte("null")
+
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface.
+// Adheres to the convention where []byte("null") is a no-op.
+func (d *Decimal) UnmarshalJSON(b []byte) error {
```
mjibson: Let's use the same method as the big.Int type: https://github.com/golang/go/blob/9d17e175e07a32f919e122faf617e09f9c9726da/src/math/big/intmarsh.go#L62

This also means we can remove the JSON methods completely since they can use the Text methods just fine. We also need both marshal and unmarshal.

decimal_test.go
```diff
@@ -743,3 +743,101 @@ func TestSizeof(t *testing.T) {
 		t.Errorf("sizeof(Context) changed: %d", s)
 	}
 }
+
+func TestUnmarshalJSON(t *testing.T) {
```
mjibson: Let's copy the test structure from math/big too: https://github.com/golang/go/blob/master/src/math/big/intmarsh_test.go#L77

Except use CmpTotal instead of Cmp.

dhui:
Thanks for the quick review @mjibson!
I've made the changes you've requested.

Note: I rebased and reset the author (didn't want to use my work email)

decimal.go
```diff
@@ -800,6 +812,30 @@ func (nd *NullDecimal) Scan(value interface{}) error {
...
+	nd.Valid = true
+	return nil
+}
+
+// MarshalText implements the encoding.TextMarshaler interface.
+func (nd *NullDecimal) MarshalText() ([]byte, error) {
+	if nd.Valid {
+		return nd.Decimal.MarshalText()
+	}
+	return []byte("null"), nil
```
dhui: @mjibson It's unclear to me what the proper string to use for marshaling null decimals.

We can support multiple values when unmarshling, but IDK of a clean way to support configurable marshling of multiple null values since the encoding.TextMarshaler doesn't provide one.

dhui:
Thanks for the quick review @mjibson!
I've made the changes you've requested.

Note: I rebased and reset the author (didn't want to use my work email)

decimal.go
dhui: Ugh... It's been a while since I've used github's review system. This comment keeps disappearing...

@mjibson It's unclear to me what the correct encoding of `null` is for `NullDecimal`. `null` will work for JSON, but it may not for other text encodings. It's easy enough to support multiple encodings of `null` in `UnmarshalText()`, but `encoding.TextMarshaler` doesn't provide an interface for configuring encodings for `MarshalText()`

decimal.go
```diff
@@ -773,6 +774,17 @@ func (d *Decimal) Scan(src interface{}) error {
...
 
+// UnmarshalText implements the encoding.TextUnmarshaler interface.
+func (d *Decimal) UnmarshalText(b []byte) error {
+	_, _, err := d.SetString(string(b))
+	return err
+}
+
+// MarshalText implements the encoding.TextMarshaler interface.
+func (d *Decimal) MarshalText() ([]byte, error) {
+	return []byte(d.String()), nil
```
mjibson: This needs to check for d == nil. See: https://github.com/golang/go/blob/c195deb48cf09970c6bb77fad96926e0af102dcf/src/math/big/intmarsh.go#L47

decimal.go
mjibson: I agree this is tricky. I think for now we should remove these methods from the NullDecimal type to force the user to decide how to handle this situation, or at least until this project gets more information about what users expect.

dhui: What do you think about implementing the following possible combinations of encoding interfaces on `NullDecimal`:
1. `json.Unmarshaler` and `json.Marshaler`
    * I'm a fan of this approach since it implements complementary interfaces and allows users to use `NullDecimal` with JSON
1. `encoding.TextUnmarshaler` and `json.Marshaler`
    * Supports generic text unmarshaling, but only supports marshaling to JSON

decimal.go
dhui: What's the expected behavior of `d == nil`?
I was thinking about returning  a nil as `[]byte` and an error but not sure if you had other thoughts. Any preferences on the error message if we go that route?

mjibson: We should do the same as in that link:

```
	if x == nil {
		return []byte("<nil>"), nil
	}
```

decimal.go
mjibson: How do we distinguish between a JSON null (not present, which shouldn't change `d` at all) and a SQL null, which should set `d` to the not valid state? That's why I think we shouldn't implement anything on NullDecimal.

dhui: I was under the impression that if the value of `NullDecimal.Valid` is `false` then the proper marshaled JSON value would be `null`. This works out if we implement the `json.Marshaler` interface (e.g. the `MarshalJSON()` method) instead of the `encode.TextMarshaler` interface (e.g. the `MarshalText()` method)

decimal.go
dhui: We'll have the same problem of determining the correct generic marshaled text value for null decimals. Maybe we should only implement the JSON marshal/unmarshal interfaces instead of the generic text ones.

mjibson: Ok I'm changing my mind. I now think we should implement both the text marshaling interfaces on both types, and not the json marshaling interfaces. The marshaling funcs will return `"<nil>"` if d is nil. NullDecimal will return `"null"` if not valid. Both will return the string value otherwise. The unmarshling funcs don't try to detect `"<nil>"` (same as math/big.Int), but instead call just SetString. However the NullDecimal unmarshal func first checks if the string is `"null"`, and if so, sets the value to null. Can you see any problems with this?

dhui: Your proposal works for me and my use case (I'm only working with JSON for now).

I understand the value of implementing the text marshaling interfaces instead of the json ones. It's nice to have a single serialization for all text formats (JSON, XML, CSV)

However I'm concerned that there's no clear generic null value. When marshaling to XML or CSV, users may want to use values other than `"null"` for interoperability with other (probably existing) systems e.g. `""`, `"nil"`, `"N/A"`, etc
We could document the behavior you proposed above to correctly set expectations with users and/or make the marshaled `nil` Decimal value configurable.

I'm fine w/ implementing the text marshaling interfaces for now and adding the appropriate docs. We can make the text marshaled `nil` value configurable later

mjibson: Yes, I agree that's a concern. I'm still going back and forth about this, but essentially that's why I proposed just not implementing anything for the NullDecimal type, which was added just for SQL compat. That would force users to encode nulls how they needed to.

dhui: I see value in `NullDecimal` supporting json encoding.
One use case is if you're consuming an external json API with null decimal values, it's nice to be able to unmarshal the json response directly to a `NullDecimal`.

Do you recommendation unmarshaling into a `*Decimal` instead?
If so, we can drop text marshaling/unmarshaling support for `NullDecimal` and have `Decimal.MarshalText()` return `<nil>` when `*d` is `nil`

mjibson: My preference is to add these methods only to the Decimal type and leave NullDecimal alone for now. Adding only the Marshal method without a corresponding Unmarshal method on NullDecimal feels like a bad design, and so I think it should be skipped until we have better reasons for making a correct choice.

dhui: I've pushed changes addresses all outstanding issues that I'm aware of. Thanks again for being open minded and collaborative!

decimal_test.go
```diff
@@ -743,3 +744,100 @@ func TestSizeof(t *testing.T) {
...
+}
+
+var encodingTests = []string{
+	"0",
+	"1",
+	"2",
+	"10",
+	"1000",
+	"1234567890",
+	"298472983472983471903246121093472394872319615612417471234712061",
```
mjibson: Add the following tests:

```
"0.0",
"NaN",
"Inf",
"123.456",
"1E1",
"1E-1",
"1.2E3",
```

decimal_test.go
```diff
@@ -743,3 +744,100 @@ func TestSizeof(t *testing.T) {
 		t.Errorf("sizeof(Context) changed: %d", s)
 	}
 }
+
+func TestUnmarshalText(t *testing.T) {
```
mjibson: math/big doesn't even test this stuff. Let's just remove this function and the one below. It's just a test of SetString at this point, which is tested elsewhere.

decimal_test.go
```diff
@@ -743,3 +744,100 @@ func TestSizeof(t *testing.T) {
...
+			}
+			s := string(b)
+			if s != tc.expectedS {
+				t.Errorf("MarshalText() test failed. %s != %s", s, tc.expectedS)
+			}
+		})
+	}
+}
+
+var encodingTests = []string{
```
mjibson: This can be moved inside TestJSONEncoding since that's the only place it's used.

decimal_test.go
dhui: `TestMarshalText()` tests `MarshalText()` with a `nil` `*Decimal`. Do you want me to keep `TestMarshalText()`, add this case as it's own test, or drop it completely?

mjibson: Remove TestUnmarshalText and TestMarshalText completely. This matches math/big.Int, and I'm ok with that.

mjibson:
Thanks for this contribution. Sorry about all the back and forth, but I think we ended up with a safe addition. If people want this for NullDecimal, we can start taking some user stories and use cases before committing to anything.

PR #68, opened 2017-09-07
Implement json.Unmarshaler and json.Marshaler interfaces

dhui:
Implementing the `encoding.TextUnmarshaler` and `encoding.TextMarshaler` interfaces aren't sufficient for decoding/encoding JSON numbers since JSON numbers aren't strings.

See the docs:
* https://golang.org/pkg/encoding/json/#Unmarshal
  > Otherwise, if the value implements encoding.TextUnmarshaler and the input is a JSON quoted string, Unmarshal calls that value's UnmarshalText method with the unquoted form of the string. 
* https://golang.org/pkg/encoding/json/#Marshal
  > If no MarshalJSON method is present but the value implements encoding.TextMarshaler instead, Marshal calls its MarshalText method and encodes the result as a JSON string.

In other words:
  - `UnmarshalText()` will support unmarshaling Decimals in JSON objects like `{"decimal": "1.1"}`
  - `UnmarshalJSON()` will support unmarshling Decimals in JSON ojbects like `{"decimal": 1.1}`

Since most JSON objects use JSON numbers instead of strings, we need to implement the `json.Unmarshaler` interface.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -15,8 +15,8 @@
 package apd
 
 import (
+	"bytes"
 	"database/sql/driver"
-	"encoding/json"
 	"math"
 	"math/big"
 	"strconv"
@@ -774,16 +774,14 @@ func (d *Decimal) Scan(src interface{}) error {
 	}
 }
 
+var jsonNull = []byte("null")
+
 // UnmarshalJSON implements the json.Unmarshaler interface.
 func (d *Decimal) UnmarshalJSON(b []byte) error {
-	var n json.Number
-	if err := json.Unmarshal(b, &n); err != nil {
-		return err
+	if bytes.Compare(b, jsonNull) == 0 {
+		return nil
 	}
-	if _, _, err := d.SetString(n.String()); err != nil {
-		return err
-	}
-	return nil
+	return d.UnmarshalText(b)
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
@@ -794,7 +792,7 @@ func (d *Decimal) UnmarshalText(b []byte) error {
 
 // MarshalJSON implements the json.Marshaler interface.
 func (d *Decimal) MarshalJSON() ([]byte, error) {
-	return []byte(d.String()), nil
+	return d.MarshalText()
 }
 
 // MarshalText implements the encoding.TextMarshaler interface.
--- a/decimal.go
+++ b/decimal.go
@@ -15,7 +15,6 @@
 package apd
 
 import (
-	"bytes"
 	"database/sql/driver"
 	"math"
 	"math/big"
@@ -774,11 +773,11 @@ func (d *Decimal) Scan(src interface{}) error {
 	}
 }
 
-var jsonNull = []byte("null")
+const jsonNull string = "null"
 
 // UnmarshalJSON implements the json.Unmarshaler interface.
 func (d *Decimal) UnmarshalJSON(b []byte) error {
-	if bytes.Compare(b, jsonNull) == 0 {
+	if string(b) == jsonNull {
 		return nil
 	}
 	return d.UnmarshalText(b)
--- a/decimal.go
+++ b/decimal.go
@@ -16,6 +16,7 @@ package apd
 
 import (
 	"database/sql/driver"
+	"encoding/json"
 	"math"
 	"math/big"
 	"strconv"
@@ -773,10 +774,27 @@ func (d *Decimal) Scan(src interface{}) error {
 	}
 }
 
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (d *Decimal) UnmarshalJSON(b []byte) error {
+	var n json.Number
+	if err := json.Unmarshal(b, &n); err != nil {
+		return err
+	}
+	if _, _, err := d.SetString(n.String()); err != nil {
+		return err
+	}
+	return nil
+}
+
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
 func (d *Decimal) UnmarshalText(b []byte) error {
 	_, _, err := d.SetString(string(b))
 	return err
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (d *Decimal) MarshalJSON() ([]byte, error) {
+	return []byte(d.String()), nil
 }
 
 // MarshalText implements the encoding.TextMarshaler interface.
--- a/decimal_test.go
+++ b/decimal_test.go
@@ -16,6 +16,7 @@ package apd
 
 import (
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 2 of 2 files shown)


decimal.go
```diff
@@ -773,12 +774,27 @@ func (d *Decimal) Scan(src interface{}) error {
 	}
 }
 
+var jsonNull = []byte("null")
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (d *Decimal) UnmarshalJSON(b []byte) error {
+	if bytes.Compare(b, jsonNull) == 0 {
```
mjibson: Let's leave this as `if string(text) == "null"` like it is for math/big. I think the compiler is smart enough to optimize it.

decimal_test.go
```diff
@@ -745,7 +745,7 @@ func TestSizeof(t *testing.T) {
 	}
 }
 
-func TestJSONEncoding(t *testing.T) {
```
mjibson: Leave this entire test exactly as it was.

decimal_test.go
```diff
@@ -768,19 +768,74 @@ func TestJSONEncoding(t *testing.T) {
...
 			}
 			if rx.CmpTotal(&tx) != 0 {
-				t.Errorf("JSON encoding of %s failed: got %s want %s", &tx, &rx, &tx)
+				t.Errorf("text encoding of %s failed: got %s want %s", &tx, &rx, &tx)
 			}
 		}
 	}
 }
+
+func TestJSONUnmarshaling(t *testing.T) {
```
mjibson: Instead of these two tests, let's copy https://github.com/golang/go/blob/d0285161f0efb1837ba0942263c33e3d53f59ece/src/math/big/intmarsh_test.go#L100 (but with the list in TestJSONEncoding, which will have to be put at the top level) because it tests the text stuff and is a good pattern from the standard library.

decimal.go
dhui: AFAIK, this will result in a memory copy every time since strings are immutable but bytes slices are mutable. The incoming `[]bytes` aren't static (depends on the JSON being unmarshaled) so the compiler can't optimize this.

decimal_test.go
dhui: Doing so will drop test coverage for `UnmarshalText()` and `MarshalText()`

dhui: Oh, ok, I understand now. We'll use similar tests for testing marshaling/unmarshling text and JSON.

decimal.go
mjibson: Please change it to the string version. Here's a few reasons:

- The compiler is pretty smart and if I had to bet, I would say that it's smart enough to avoid the string -> bytes copy each call.
- The function call into bytes.Compare isn't cheap.
- The go stdlib already does this, and I have no reason to not copy them.
- My benchmarks show that the string version is faster:

```
func BenchmarkBytes(b *testing.B) {
	var bb = []byte("blah")
	var a bool
	for i := 0; i < b.N; i++ {
		a = bytes.Compare(bb, bb) == 0
	}
	if !a {
		b.Fatal("balh")
	}
}

func BenchmarkString(b *testing.B) {
	var xb = []byte("blah")
	var a bool
	for i := 0; i < b.N; i++ {
		a = string(xb) == "blah"
	}
	if !a {
		b.Fatal("balh")
	}
}
```

```

BenchmarkBytes-12     	500000000	         3.14 ns/op
BenchmarkString-12    	2000000000	         0.62 ns/op
```

My mind could be changed by any evidence, like a benchmark or assembly output analysis that show differently, but as it is there's no evidence that the bytes version is better.

decimal_test.go
```diff
@@ -755,8 +756,6 @@ func TestJSONEncoding(t *testing.T) {
 		"1234567890",
 		"298472983472983471903246121093472394872319615612417471234712061",
 		"0.0",
-		"NaN",
```
mjibson: What are the test failures with these?

decimal_test.go
```diff
@@ -779,7 +778,47 @@ func TestJSONEncoding(t *testing.T) {
...
 			if rx.CmpTotal(&tx) != 0 {
-				t.Errorf("JSON encoding of %s failed: got %s want %s", &tx, &rx, &tx)
+				t.Errorf("text encoding of %s failed: got %s want %s", &tx, &rx, &tx)
+			}
+		}
+	}
+}
+
+func TestXMLEncoding(t *testing.T) {
+	var encodingTests = []string{
```
mjibson: Use the encodingTests from the other test: move it to the top level.

decimal_test.go
dhui: `Nan` and `Inf` aren't valid JSON numbers.
http://www.json.org/
http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

decimal.go
dhui: Odd, I get similar benchmark performances even with the following changes:
1. Use `bytes.Equal()` instead of `bytes.Compare`
1. Use a table of byte slices to force string copying conversion

```
package comparison

import (
	"bytes"
	"math/rand"
	"testing"
	"time"
)

var sampleBytes = [][]byte{
	[]byte("blah"),
	[]byte(""),
	[]byte("null"),
	[]byte("foo"),
	[]byte("bar"),
	[]byte("foobar"),
	[]byte("hello"),
	[]byte("world"),
	[]byte("helloworld"),
	[]byte("deadbeef"),
}

const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
const (
	letterIdxBits = 6                    // 6 bits to represent a letter index
	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
)

var src = rand.NewSource(time.Now().UnixNano())

func RandBytes(n int) []byte {
	b := make([]byte, n)
	// A src.Int63() generates 63 random bits, enough for letterIdxMax characters!
	for i, cache, remain := n-1, src.Int63(), letterIdxMax; i >= 0; {
		if remain == 0 {
			cache, remain = src.Int63(), letterIdxMax
		}
		if idx := int(cache & letterIdxMask); idx < len(letterBytes) {
			b[i] = letterBytes[idx]
			i--
		}
		cache >>= letterIdxBits
		remain--
	}
	return b
}

func init() {
	for i := 0; i < 100; i++ {
		sampleBytes = append(sampleBytes, RandBytes(100))
	}
}

func BenchmarkBytes(b *testing.B) {
	var bb = []byte("blah")
	var a bool
	for i := 0; i < b.N; i++ {
		j := i % len(sampleBytes)
		a = bytes.Equal(bb, sampleBytes[j])
	}
	b.Log(a)
}

func BenchmarkString(b *testing.B) {
	var xb = []byte("blah")
	var a bool
	for i := 0; i < b.N; i++ {
		j := i % len(sampleBytes)
		a = string(xb) == string(sampleBytes[j])
	}
	b.Log(a)
}
```
```
goos: darwin
goarch: amd64
BenchmarkBytes-8    	100000000	        13.2 ns/op
--- BENCH: BenchmarkBytes-8
	comparison_test.go:62: true
	comparison_test.go:62: false
	comparison_test.go:62: false
	comparison_test.go:62: false
	comparison_test.go:62: false
BenchmarkString-8   	100000000	        10.4 ns/op
--- BENCH: BenchmarkString-8
	comparison_test.go:72: true
	comparison_test.go:72: false
	comparison_test.go:72: false
	comparison_test.go:72: false
	comparison_test.go:72: false
PASS
```

So I'll change to the string conversion.

decimal_test.go
dhui: Do you want me to pull out the common set to the top level and add the `Nan` and `Inf` cases for xml/text?

mjibson:
I was curious about math/big.Float, and it appears to not support being a json number. See: https://play.golang.org/p/Huh7aeKc95

So I'm now very hesitant to merge this PR. Why is apd different than big.Float?

mjibson:
The nice thing about leaving it as a quoted string in JSON is that nan and inf would be supported correctly. I'm really not ok with having those values just be not supported. If we did support the json number format, then we'd have to error on those values. So overall I think keeping them strings is correct.

dhui:
That makes sense. Feel free to close this PR then.
I do what I was doing before. e.g. embed the `apd.Decimal` struct in my own struct and implement the `json.Marshaler` and `json.Unmarshaler` interfaces.

mjibson:
Ok. But I really appreciate bringing this issue up and working on the code for this. Sorry we decided to not merge it, but the added knowledge about *why* we aren't merging it is very useful to us for the future.

dhui:
You're welcome and thanks for considering the change and taking the time to review it.

Agreed about the learnings!

PR #69, opened 2017-09-08
Improve docs for CmpTotal

### LINKED ISSUE #66: Improve docs for cmp, cmptotal
It is not obvious to users that CmpTotal doesn't compare numeric values. See #65

mjibson:
Fixes #66

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -412,16 +412,21 @@ func (d *Decimal) setBig(b *big.Int) *big.Int {
 	return b
 }
 
-// CmpTotal compares d and x and returns:
+// CmpTotal compares d and x using their abstract representation rather
+// than their numerical value. A total ordering is defined for all possible
+// abstract representations, as described below. If the first operand is
+// lower in the total order than the second operand then the result is -1,
+// if the operands have the same abstract representation then the result is
+// 0, and if the first operand is higher in the total order than the second
+// operand then the result is 1.
 //
-//   -1 if d <  x
-//    0 if d == x
-//   +1 if d >  x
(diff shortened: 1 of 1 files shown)


decimal.go
```diff
@@ -412,14 +412,19 @@ func (d *Decimal) setBig(b *big.Int) *big.Int {
 	return b
 }
 
-// CmpTotal compares d and x and returns:
+// CmpTotal compares d and x and compares them using their abstract
```
nvanbenschoten: > compares d and x and compares them

The wording is off here

decimal.go
```diff
@@ -412,14 +412,19 @@ func (d *Decimal) setBig(b *big.Int) *big.Int {
...
-//   +1 if d >  x
-//
-// This comparison uses a total ordering that is defined to compare all
-// finite and non-finite (special) values.
+// Numbers (representations which are not NaNs) are ordered such that a
+// larger numerical value is higher in the ordering. If two representations
+// have the same numerical value then the exponent is taken into account;
+// larger (more positive) exponents are higher in the ordering.
 //
 // For example, the following values are ordered from lowest to highest:
```
nvanbenschoten: Perhaps add in the sentence: `Note the difference in ordering between 1.2300 and 1.23.`. That's the key example.

mjibson:
I've linked to that site once at the top of the docs, and I think that's sufficient. In general I want the docs here to be self-contained, because that spec is too verbose for most users to care about or understand. (It took me weeks of working with it before I understood how it worked together.) So I don't think a link in this function is worth it.

Issue #22, opened 2017-02-16
Investigate Newton–Raphson for Quo


mjibson:
See https://github.com/cockroachdb/apd/issues/13#issuecomment-280407542

mjibson:
This is difficult because this method uses successive approximation, which makes it difficult to determine exactness. That is, the current long division method maintains a remainder which can be used to know if the division was exact or not. I'm not sure how Newton-Raphson could correctly report exactness.

@eisenstatdavid

eisenstatdavid:
I'll take a crack at this when I have time.

mjibson:
I tried real hard to make this work but couldn't. Some code linked in those PRs.

Issue #13, opened 2017-02-10
Quo needs a review


mjibson:
spec at: http://speleotrove.com/decimal/daops.html#refdivide

eisenstatdavid:
Does this include `QuoInteger`?

eisenstatdavid:
Both LGTM, subject to two hopefully expected caveats:

1. Special values are not handled properly. ISTM that the decimal type can't represent NaN, -∞, -0, or ∞.
2. `Quo` implements long division, which is slow for large operands. [Newton–Raphson division](https://en.wikipedia.org/wiki/Division_algorithm) is one possible upgrade path.

mjibson:
No, QuoInteger isn't required. It's a much simpler algorithm.

I wasn't aware of Newton-Raphson. I'll open an issue so we don't forget.

PR #72, opened 2017-09-14
[DNM] implement Quo using Newton-Raphson

mjibson:
In attempting to work on #22, I implemented Newton-Raphson. I got around the exactness problem the same way cuberoot does: multiply the result by the input and see if it's an exact match. All of the tests pass with this code.

However I was not able to observe any performance increase. Instead, I saw a huge performance decrease:

```
name                old time/op  new time/op     delta
GDA/divide-8        1.49ms ± 1%     9.21ms ± 1%    +517.48%  (p=0.008 n=5+5)
GDA/divideint-8      373µs ± 0%      393µs ± 1%      +5.45%  (p=0.008 n=5+5)
GDA/exp-8            262ms ± 2%      339ms ± 3%     +29.02%  (p=0.008 n=5+5)
GDA/ln-8             162ms ± 2%      291ms ± 2%     +79.24%  (p=0.008 n=5+5)
GDA/log10-8          206ms ± 3%      362ms ± 1%     +75.51%  (p=0.008 n=5+5)
GDA/power-8          425ms ± 1%      906ms ± 1%    +113.30%  (p=0.008 n=5+5)
GDA/powersqrt-8      889ms ± 1%     2066ms ± 2%    +132.34%  (p=0.008 n=5+5)
GDA/randoms-8       8.57ms ± 0%  1354.78ms ± 1%  +15701.80%  (p=0.016 n=4+5)
GDA/squareroot-8    71.9ms ± 1%    213.9ms ± 1%    +197.48%  (p=0.008 n=5+5)
GDA/cuberoot-apd-8  4.18ms ± 0%     9.89ms ± 1%    +136.57%  (p=0.008 n=5+5)
```

It was mentioned in #13 that this algorithm is fast for large operands. I also performed a number of benchmarks with very large (>100 digits) operands in just the numerator, denominator, and both. In no case was I ever able to produce a benchmark that was faster using this method.

Some ideas about why this is: the first part reduces the denominator to [0.5, 1]. Normally this can be done very quickly by simply bit shifting as many times as needed if you are in base 2. However, being in base 10, we cannot do that, and must instead multiply the inputs each time. Ok I just did some benchmarks where the input is already within that range and it's still slower in the various cases I tested.

Another possibility is that this method has to use a precision 3 greater than that needed to have enough precision to round correctly. I really don't think it's this but who knows.

A third idea is the use of the loop convergence. Newton-Raphson is able to compute the number of iterations needed: maybe the loop is doing a few too many? My guess is this isn't a problem.

Anyone have other ideas or see things I'm doing clearly wrong? I used the psuedocode from wikipedia; https://en.wikipedia.org/wiki/Division_algorithm#Pseudocode.

@eisenstatdavid @RaduBerinde tagging you in case you're interested in this since ya'll are into algorithms.


RaduBerinde:
I don't see the problem with the code, it follows the pseudocode AFAICT. I would add some printfs that show all iterations of the loop to see if we are doing too many iterations, and if the number of iterations is similar to what they say `(log2((P+1)/log2(17))` (note that `P` is binary precision though)

I would search for other adaptations of the algorithm that are tailored to base-10 decimal (the pseudocode as written seems to be done with base-2 floating-point in mind).

---

---

*[context.go, line 308 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/72#-Ku-PosGmhSPqD_6TdGS:-Ku-PosGmhSPqD_6TdGT:bq8vlmq) ([raw file](https://github.com/cockroachdb/apd/blob/d948619898b5337d08258ec0468ef67df3df7d71/context.go#L308)):*
> ```Go
> 	D.Negative = false
> 	exp2 := 0
> 	for D.Cmp(decimalHalf) < 0 {
> ```

I would also check if these loops aren't taking a lot of time (we could look at multiplying/dividing by `2^something` first; or, we could calculate an initial exponent value from the base-10 exponent of D)

---

*[context.go, line 325 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/72#-Ku-OYywZmDj6MDJcbJx:-Ku-OYywZmDj6MDJcbJy:b-9dg4mc) ([raw file](https://github.com/cockroachdb/apd/blob/d948619898b5337d08258ec0468ef67df3df7d71/context.go#L325)):*
> ```Go
> 	X0 := new(Decimal)
> 	for loop := nc.newLoop("quo", X, c.Precision, 1); ; {
> 		// X = X + X * (1 - D * X)
> ```

Why not `X = X * (2 - D * X)`? My guess is that the pseudocode we're following is written with floating-point in mind, it might make more sense in that context (?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/72#-:-Ku-P9BSqAkd0GqfjQkn:bva2gzk)*

mjibson:

---

*[context.go, line 308 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/72#-Ku-PosGmhSPqD_6TdGS:-Ku3WSw1yeNQ4J4P0_0O:b-6nvfz7) ([raw file](https://github.com/cockroachdb/apd/blob/d948619898b5337d08258ec0468ef67df3df7d71/context.go#L308)):*

I would also check if these loops aren't taking a lot of time (we could look at multiplying/dividing by `2^something` first; or, we could calculate an initial exponent value from the base-10 exponent of D)
</blockquote></details>

I did a test about this by benchmarking inputs that were already within [0.5, 1], and it was still slower. But I do think moving by a factor of 10 would be an improvement in the general case since it's so easy to modify the exponent.

---

*[context.go, line 325 at r1](https://reviewable.io:443/reviews/cockroachdb/apd/72#-Ku-OYywZmDj6MDJcbJx:-Ku3Wg-eTilnuHIN9DRZ:bskngl3) ([raw file](https://github.com/cockroachdb/apd/blob/d948619898b5337d08258ec0468ef67df3df7d71/context.go#L325)):*
The wikipedia article addresses this:

> From a computation point of view, the expressions {\displaystyle X_{i+1}=X_{i}+X_{i}(1-DX_{i})} X_{i+1}=X_{i}+X_{i}(1-DX_{i}) and {\displaystyle X_{i+1}=X_{i}(2-DX_{i})} X_{i+1}=X_{i}(2-DX_{i}) are not equivalent. To obtain a result with a precision of n bits while making use of the second expression, one must compute the product between {\displaystyle X_{i}} X_{i} and {\displaystyle (2-DX_{i})} (2-DX_{i}) with double the required precision (2n bits). In contrast, the product between {\displaystyle X_{i}} X_{i} and {\displaystyle (1-DX_{i})} (1-DX_{i}) need only be computed with a precision of n bits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/apd/72)*

mjibson:
As a test (https://github.com/cockroachdb/apd/commit/9af50e51015859ee380440bc1d136c4889c2cd6d) I implemented Goldschmidt division using the binomial simplification (https://en.wikipedia.org/wiki/Division_algorithm#Binomial_theorem). It performed slightly better than (about 7% faster) Newton-Raphson, but still real slow:

```
name \ time/op   old.txt     newton.txt   goldschmidt.txt
GDA/divide-4     887µs ±16%  6672µs ± 2%      6171µs ± 3%
GDA/divideint-4  120µs ±11%   140µs ±25%       141µs ±19%
```

It has one less operation per loop (a Sub), but the same number of multiplications. In order to pass the tests it has to bump the precision up pretty high (`c.Precision*2+1`). I don't have a lot of ideas about how to make this 10x faster to compete with what we have now, so unless someone else wants to take a stab, I'm going to close the issue in a few days.

PR #81, opened 2018-10-17
Final cleanup before next release

mjibson:


danhhz:
Ha! I'm surprised you've never run into "sciencedog" here. It's cockroach jargon (as far as i know, i've never seen it anywhere else) for admitting you're not qualified to sign off on something

PR #82, opened 2018-12-05
add BSON marshalling

reddec:
Add BSON marshal and unmarshal

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,6 @@
 module github.com/cockroachdb/apd
 
-require github.com/pkg/errors v0.8.0
+require (
+	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
+	github.com/pkg/errors v0.8.0
+)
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,3 +1,6 @@
 module github.com/cockroachdb/apd
 
-require github.com/pkg/errors v0.8.0
+require (
+	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
```
mjibson: I don't think adding this dependency is worth the PR here, so I'm unlikely to merge this.

reddec: Yes, it's may looks strange but I didn't find any better way to do it. Any other ideas?

mjibson: You'll need to implement this in another package sadly. Like make some decimal wrapper type somewhere else that implements the bson methods. Unless there's very strong motivation it doesn't make sense to add them here.

reddec: Ok, that's make sense. Using additional wrapper may cause additional problems during integration to a existing code base.
Unfortunately looks like I should maintain my fork separately merging changes from here if you are not against.
However if for some reason you will change your decision I will be happy to make a PR again

mjibson: Yes, maintaining your own fork is definitely the best way to go for now. Also note that we decided against even supporting JSON (https://github.com/cockroachdb/apd/pull/68#issuecomment-328457313), so a format not in the stdlib has an even higher bar of getting in.

PR #88, opened 2019-04-06
apd: add decomposer interface methods

kardianos:
This implements the interface as worked on in the
issue https://golang.org/issue/30870 .

### DIFF
--- a/decomposer.go
+++ b/decomposer.go
@@ -1,0 +1,116 @@
+// Copyright 2016 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package apd
+
+import (
+	"fmt"
+)
+
+// decomposer composes or decomposes a decimal value to and from individual parts.
+// There are four separate parts: a boolean negative flag, a form byte with three possible states
+// (finite=0, infinite=1, NaN=2),  a base-2 little-endian integer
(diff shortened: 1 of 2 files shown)


kardianos:
FYI: I think I have a crdb CLA signed, but I if it isn't on file (I think the CLA service they were using at first went down), I can sign a new one.

Intent is to post the PRs as drafts on a few decimal packages, then if we can agree on the interface, commit them around the same time. The interface is designed to be self-contained, not needing to reference any outside package, but I posted a package for human reference here: https://github.com/golang-sql/decomposer .

kardianos:
@mjibson Would you be willing to consider this PR?

decomposer.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil
+	}
+	// Finite form.
+	d.Negative = negative
+
+	// coefficient is a little-endian, swap for big-endian.
+	i := 0
+	j := len(coefficient) - 1
+	for i < j {
+		coefficient[i], coefficient[j] = coefficient[j], coefficient[i]
```
mjibson: Is mutating coefficient ok? Feels like this API should document that if so.

kardianos: Great observation. I've updated the [documentation](https://godoc.org/github.com/golang-sql/decomposer#Decimal) to say it is *not* okay to modify the coefficient and updated the implementation here to make a copy first and test for this case.

decomposer.go
```diff
@@ -0,0 +1,121 @@
...
+		d.Form = NaN
+		d.Negative = negative
+		return nil
+	}
+	// Finite form.
+	d.Negative = negative
+
+	// coefficient is a little-endian, swap for big-endian.
+	c2 := make([]byte, len(coefficient))
+	copy(c2, coefficient)
```
mjibson: This copy is unneeded since a copy is performed anyway. I think this is better, and moves i and j into the for loop scope.

```
	c2 := make([]byte, len(coefficient))
	for i, j := 0, len(c2) - 1; i < j; {
		c2[i], c2[j] = coefficient[j], coefficient[i]
		i++
		j--
	}
```

kardianos: Yep, you're correct again. Sorry for not seeing that. The suggested code was close, but still failed unit tests. I did a simpler thing and just copied all bytes in reverse.

PR #97, opened 2020-10-08
apd: add Size method

yuzefovich:


asubiotto:
It doesn't feel like this is something that `apd` should provide. We already have a `SizeOfDecimal` function in `pkg/sql/sem/tree/datum.go`, is that not what you need?

yuzefovich:
Oh, indeed, I don't why I didn't check that myself.

PR #101, opened 2022-01-01
[DNM] apd: embed small coefficient values in Decimal struct

nvanbenschoten:
Replaces https://github.com/cockroachdb/cockroach/pull/74369.

This commit introduces a performance optimization that embeds small coefficient values directly in their `Decimal` struct, instead of storing these values in a separate heap allocation.

Each `Decimal` maintains (through `big.Int`) an internal reference to a variable-length coefficient, which is represented by a `[]big.Word`. The `coeffInline` field and the `lazyInit` method combine to allow us to inline this coefficient within the `Decimal` struct when its value is sufficiently small. In `lazyInit`, we point the `Coeff` field's backing slice at the `coeffInline` array. `big.Int` will avoid re-allocating this array until it is provided with a value that exceeds the initial capacity. We set this initial capacity to accommodate any coefficient that would fit in a 64-bit integer (i.e. values up to 2^64).

This is an alternative to an optimization that many other arbitrary precision decimal libraries have where small coefficient values are stored as numeric fields in their data type's struct. Only when this coefficient value gets large enough do these libraries fall back to a variable-length coefficient with internal indirection. We can see the optimization in practice in the `ericlagergren/decimal` library, where each struct contains a `compact uint64` and an `unscaled big.Int`. Prior concern from the authors of `cockroachdb/apd` regarding this form of compact representation optimization was that it traded performance for complexity. The optimization fractured control flow, leaking out across the library and leading to more complex, error-prone code.

The approach taken in this commit does not have the same issue. All arithmetic on the decimal's coefficient is still deferred to `bit.Int`. In fact, the entire optimization is best-effort, and bugs that lead to missed calls to `lazyInit` are merely missed opportunities to avoid a heap allocation, and nothing more serious. The internal reference within Decimal is vulnerable to memory aliasing bugs when a `Decimal` is copied by value (and not through the `Decimal.Set` method), but this is not a new concern.

### Impact on benchmarks:

```
name                   old time/op    new time/op    delta
GDA/compare-10           61.5µs ±20%    54.5µs ±18%  -11.34%  (p=0.035 n=10+10)
GDA/reduce-10            18.8µs ± 7%    17.5µs ± 6%   -7.16%  (p=0.001 n=10+10)
GDA/remainder-10         67.0µs ±11%    62.2µs ± 6%   -7.09%  (p=0.023 n=10+10)
Exp/P5/S-4/D-2-10        2.56µs ± 4%    2.51µs ± 4%     ~     (p=0.060 n=10+10)
[...]

PR #102, opened 2022-01-03
[DNM] apd: embed small coefficient values in Decimal struct

nvanbenschoten:
Replaces cockroachdb/cockroach#74369.
Replaces https://github.com/cockroachdb/apd/pull/101.

This commit introduces a performance optimization that embeds small coefficient values directly in their `Decimal` struct, instead of storing these values in a separate heap allocation. It does so by replacing `math/big.Int` with a new wrapper type called `BigInt` that provides an "inline" compact representation optimization.

Each `BigInt` maintains (through `big.Int`) an internal reference to a variable-length integer value, which is represented by a `[]big.Word`. The `_inline` field and `lazyInit` method combine to allow `BigInt` to inline this variable-length integer array within the `BigInt` struct when its value is sufficiently small. In `lazyInit`, we point the `_inner` field's slice at the `_inline` array. `big.Int` will avoid re-allocating this array until it is provided with a value that exceeds the initial capacity.

We set the capacity of the inline array to accommodate any value that would fit in a 128-bit integer (i.e. values up to 2^128 - 1).

This is an alternative to an optimization that many other arbitrary precision decimal libraries have where small coefficient values are stored as numeric fields in their data type's struct. Only when this coefficient value gets large enough do these libraries fall back to a variable-length coefficient with internal indirection. We can see the optimization in practice in the `ericlagergren/decimal` library, where each struct contains a `compact uint64` and an `unscaled big.Int`. Prior concern from the authors of `cockroachdb/apd` regarding this form of compact representation optimization was that it traded performance for complexity. The optimization fractured control flow, leaking out across the library and leading to more complex, error-prone code.

The approach taken in this commit does not have the same issue. All arithmetic on the decimal's coefficient is still deferred to `big.Int`. In fact, the entire optimization is best-effort, and bugs that lead to missed calls to `lazyInit` are merely missed opportunities to avoid a heap allocation, and nothing more serious.

However, one major complication with this approach is that Go's escape analysis struggles with understanding self-referential pointers. A naive implementation of this optimization would force all `BigInt` structs to escape to the heap. To work around this, we employ a similar trick to `sync.Cond` and `strings.Builder`. We trick escape analysis to allow the self-referential pointer without causing the struct to escape.

This works but it introduces complexity if `BigInt` structs are copied by value. So to avoid nasty bugs, we disallow copying of `BigInt` structs. The self-referencing pointer from `_inner` to `_inline` makes this unsafe, as it could allow aliasing between two `BigInt` structs which would be hidden from escape analysis. If the first `BigInt` then fell out of scope and was GCed, this could corrupt the state of the second `BigInt`. `sync.Cond` and `strings.Builder` also prevent copying to avoid this kind of issue. In fact, `big.Int` itself says that "shallow copies are not supported and may lead to errors", but it doesn't enforce this.

### Impact on benchmarks:

```
name                 old time/op    new time/op    delta
GDA/comparetotal-10    46.3µs ± 0%    24.4µs ± 1%   -47.33%  (p=0.000 n=10+9)
GDA/remainder-10       68.4µs ± 0%    40.2µs ± 0%   -41.31%  (p=0.000 n=10+9)
GDA/abs-10             11.5µs ± 1%     7.0µs ± 0%   -39.46%  (p=0.000 n=10+10)
GDA/compare-10         55.7µs ± 0%    33.8µs ± 1%   -39.25%  (p=0.000 n=10+10)
GDA/tointegralx-10     36.0µs ± 1%    22.1µs ± 0%   -38.55%  (p=0.000 n=10+9)
GDA/minus-10           14.1µs ± 0%     8.8µs ± 0%   -38.10%  (p=0.000 n=10+10)
GDA/tointegral-10      35.1µs ± 1%    21.8µs ± 0%   -37.83%  (p=0.000 n=10+10)
GDA/quantize-10         134µs ± 1%      84µs ± 0%   -37.57%  (p=0.000 n=9+10)
GDA/subtract-10         171µs ± 0%     109µs ± 0%   -36.37%  (p=0.000 n=10+10)
GDA/reduce-10          21.7µs ± 1%    14.0µs ± 0%   -35.18%  (p=0.000 n=10+8)
GDA/divideint-10       34.2µs ± 0%    22.8µs ± 0%   -33.40%  (p=0.000 n=9+9)
GDA/multiply-10        80.5µs ± 0%    54.9µs ± 0%   -31.83%  (p=0.000 n=9+10)
GDA/randoms-10         3.20ms ± 0%    2.21ms ± 0%   -30.73%  (p=0.000 n=9+9)
GDA/add-10              917µs ± 0%     641µs ± 0%   -30.03%  (p=0.000 n=10+10)
GDA/rounding-10         623µs ± 0%     472µs ± 0%   -24.16%  (p=0.000 n=10+8)
GDA/plus-10            45.0µs ± 0%    37.5µs ± 0%   -16.63%  (p=0.000 n=10+9)
GDA/base-10             131µs ± 0%     114µs ± 0%   -13.40%  (p=0.000 n=10+10)
GDA/squareroot-10      31.6ms ± 0%    27.4ms ± 0%   -13.16%  (p=0.000 n=9+8)
GDA/powersqrt-10        431ms ± 0%     417ms ± 0%    -3.16%  (p=0.000 n=9+9)
GDA/divide-10           366µs ± 0%     360µs ± 0%    -1.72%  (p=0.000 n=9+8)
GDA/cuberoot-apd-10    1.97ms ± 0%    2.07ms ± 0%    +5.21%  (p=0.000 n=9+10)
GDA/exp-10              119ms ± 0%     126ms ± 0%    +6.19%  (p=0.000 n=10+8)
GDA/power-10            208ms ± 0%     225ms ± 0%    +8.55%  (p=0.000 n=10+9)
GDA/log10-10            101ms ± 0%     110ms ± 0%    +9.49%  (p=0.000 n=10+9)
GDA/ln-10              79.6ms ± 0%    87.3ms ± 0%    +9.61%  (p=0.000 n=9+10)

name                 old alloc/op   new alloc/op   delta
GDA/abs-10             6.50kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/compare-10         39.1kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)
GDA/comparetotal-10    37.2kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)
GDA/minus-10           7.71kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/reduce-10          10.1kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)
GDA/remainder-10       45.5kB ± 0%     0.1kB ± 0%   -99.86%  (p=0.000 n=10+10)
GDA/rounding-10         292kB ± 0%       5kB ± 0%   -98.33%  (p=0.000 n=10+10)
GDA/squareroot-10      8.46MB ± 0%    0.29MB ± 0%   -96.55%  (p=0.000 n=10+10)
GDA/randoms-10         1.25MB ± 0%    0.05MB ± 0%   -95.98%  (p=0.000 n=10+10)
GDA/divideint-10       23.2kB ± 0%     1.2kB ± 0%   -94.85%  (p=0.000 n=10+10)
GDA/divide-10           102kB ± 0%       6kB ± 0%   -93.64%  (p=0.000 n=10+10)
GDA/powersqrt-10       77.8MB ± 0%     5.1MB ± 0%   -93.44%  (p=0.000 n=10+10)
GDA/quantize-10        76.4kB ± 0%     8.9kB ± 0%   -88.33%  (p=0.000 n=8+8)
GDA/multiply-10        55.4kB ± 0%    10.7kB ± 0%   -80.71%  (p=0.000 n=10+10)
GDA/tointegralx-10     27.9kB ± 0%     6.2kB ± 0%   -77.89%  (p=0.000 n=10+10)
GDA/tointegral-10      27.2kB ± 0%     6.2kB ± 0%   -77.34%  (p=0.000 n=9+10)
GDA/subtract-10         131kB ± 0%      34kB ± 0%   -73.82%  (p=0.000 n=10+7)
GDA/cuberoot-apd-10     265kB ± 0%      82kB ± 0%   -68.97%  (p=0.000 n=10+10)
GDA/base-10            60.3kB ± 0%    26.2kB ± 0%   -56.61%  (p=0.000 n=10+10)
GDA/ln-10              10.2MB ± 0%     5.0MB ± 0%   -50.58%  (p=0.000 n=10+10)
GDA/log10-10           12.5MB ± 0%     6.4MB ± 0%   -49.20%  (p=0.000 n=10+10)
GDA/add-10              811kB ± 0%     422kB ± 0%   -47.95%  (p=0.000 n=10+10)
GDA/power-10           27.0MB ± 0%    14.3MB ± 0%   -47.21%  (p=0.000 n=10+10)
GDA/plus-10            52.3kB ± 0%    40.0kB ± 0%   -23.47%  (p=0.000 n=10+8)
GDA/exp-10             61.3MB ± 0%    55.7MB ± 0%    -9.12%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
GDA/abs-10                238 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/compare-10          1.24k ± 0%     0.00k       -100.00%  (p=0.000 n=10+10)
GDA/comparetotal-10       881 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/minus-10              274 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/reduce-10             352 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/remainder-10        1.40k ± 0%     0.00k ± 0%   -99.93%  (p=0.000 n=10+10)
GDA/quantize-10         2.41k ± 0%     0.04k ± 0%   -98.18%  (p=0.000 n=10+10)
GDA/squareroot-10        278k ± 0%        5k ± 0%   -98.09%  (p=0.000 n=10+10)
GDA/randoms-10          42.2k ± 0%      0.9k ± 0%   -97.81%  (p=0.000 n=10+10)
GDA/rounding-10         10.2k ± 0%      0.3k ± 0%   -96.83%  (p=0.000 n=10+10)
GDA/divide-10           3.08k ± 0%     0.10k ± 0%   -96.72%  (p=0.000 n=10+10)
GDA/tointegralx-10        857 ± 0%        35 ± 0%   -95.92%  (p=0.000 n=10+10)
GDA/tointegral-10         833 ± 0%        35 ± 0%   -95.80%  (p=0.000 n=10+10)
GDA/subtract-10         3.51k ± 0%     0.18k ± 0%   -94.76%  (p=0.000 n=10+10)
GDA/powersqrt-10        2.63M ± 0%     0.17M ± 0%   -93.49%  (p=0.000 n=10+10)
GDA/multiply-10         1.35k ± 0%     0.16k ± 0%   -88.41%  (p=0.000 n=10+10)
GDA/cuberoot-apd-10     7.75k ± 0%     1.19k ± 0%   -84.59%  (p=0.000 n=10+10)
GDA/add-10              15.2k ± 0%      3.0k ± 0%   -79.96%  (p=0.000 n=10+10)
GDA/divideint-10          704 ± 0%       142 ± 0%   -79.83%  (p=0.000 n=10+10)
GDA/ln-10                269k ± 0%       80k ± 0%   -70.21%  (p=0.000 n=10+10)
GDA/log10-10             327k ± 0%      101k ± 0%   -69.04%  (p=0.000 n=10+10)
GDA/power-10             700k ± 0%      226k ± 0%   -67.69%  (p=0.000 n=10+10)
GDA/plus-10               706 ± 0%       229 ± 0%   -67.56%  (p=0.000 n=10+10)
GDA/exp-10               911k ± 0%      544k ± 0%   -40.32%  (p=0.000 n=10+10)
GDA/base-10             2.86k ± 0%     2.52k ± 0%   -12.06%  (p=0.000 n=10+10)
```

cc. @mjibson


nvanbenschoten:
## Updates

### Integration into CRDB

I took a stab at integrating this into CRDB in [this nightmare of a prototype](https://github.com/nvanbenschoten/cockroach/commit/83e9dad8712b6453cccb4161804b8bf8769cc9f4).

The restriction that `apd.Decimal` can no longer be copied by value wasn't too much of a hurdle in most places. In fact, it actually forced some improvements in the key and value decoding code, as it prevented memcpys and wasted allocations by preventing us from writing code that caused them. For example, `DecodeDecimalAscending`'s signature changed from:
```
func DecodeDecimalAscending(buf []byte, tmp []byte) ([]byte, apd.Decimal, error)
```
to
```
func DecodeDecimalAscending(dec *apd.Decimal, buf []byte, tmp []byte) ([]byte, error)
```
This in turn forces the code to be slightly more deliberate about where it is decoding `Decimal`s into, so we end up using the `rowenc.DatumAlloc` more reliably. All in all, I think that's fine.

The more complex part of the change was in the vectorized execution engine. The complexity wasn't even entirely due to this change, but instead due to the need to first address [this todo](https://github.com/cockroachdb/cockroach/blob/798a3e544d7ed38c5ea21085054b97cac067f56d/pkg/col/coldata/native_types.go#L109-L110):
```
// TODO(yuzefovich): consider whether Get and Set on Decimals should operate on
// pointers to apd.Decimal.
```
Since we can no longer copy `Decimal`s by value, the handling of `Decimal` in the vectorized execution needs to become even more custom. This is because we don't currently have any data types that the vectorized execution engine interacts with by reference.

I spent a while learning about and playing with `execgen`. I made some progress, but eventually gave up and made changes manually. I'm sure someone who actually knows how `execgen` works would have more luck, but I really struggled with the conditional templating.

After making that change and getting CRDB compiling, I took it for a spin. First, I confirmed that we see the same reduction in heap allocations that we saw in https://github.com/cockroachdb/cockroach/pull/74369. Next, I played around a bit with the tpcds dataset, which includes many DECIMAL columns. I didn't run any tpcds queries, but I did do a few custom full-table aggregations over the DECIMAL columns in ~2GB large tables.

Nothing was scientific, but there did seem to be something around a 5% speedup compared to master on these aggregations. Or maybe I just wanted there to be. Again, not scientific.

----

### More compact memory representation

After writing this change I looked into how other systems like Postgres, MySQL, SQLServer, and Materialize handle Decimals. It turns out that only Postgres (not even Materialize, which strives for close to full PG support) supports arbitrary precision decimals (100k+ precision). Most other systems cap the precision of decimals at somewhere around 40, meaning that they don't even need to support a variable-length memory representation. They can instead inline the entire value in a single 24 byte large struct.

We do want to maintain compatibility with PG and so I think we need to continue to support arbitrary precision, but this made me question why we were tailoring so much of this code to arbitrary precision. Specifically, I started questioning why we were spending so many bytes of the `Decimal` memory representation to support absurdly large precision value. Out of the 64 bytes in a `Decimal`, 32 were going to the `big.Int` (bool + slice header) and 8 were going to the self-pointer to dynamically detect value copies which are unsafe as a second-order consequence of supporting arbitrary precision.

I drafted another prototype that further optimizes for small values at the expense of large values:  https://github.com/nvanbenschoten/apd/commit/654be62f74b7189b16c929bdb561f86a878d71f1.

The key idea is that instead of a memory representation that looks like:
```go
type Decimal struct {
    Form     int8
    Negative bool
    Exponent int32
    Coeff    BigInt {
        _inner big.Int {
            neg bool
            abs []uint
        }
        _inline [2]uint
        _addr   uintptr
    }
} // sizeof = 64 bytes
```

We have a memory representation that looks like:
```go
type Decimal struct {
    Form     int8
    Negative bool
    Exponent int32
    Coeff    BigInt {
        _inner *big.Int
        _inline [2]uint
    }
} // sizeof = 32
```

There are a few tradeoffs here:

Pros:
- reduces memory size from 64 bytes to 32 bytes, allowing us to pack two `Decimals` into a single cache line
- removes the self-referential pointer, making the Decimal struct copyable again

Cons:
- two heap allocations for large values instead of one (allocate `big.Int` then allocate large backing array)
- need to inflate stack-allocated `big.Int` on demand during arithmetic if we wanted to continue to defer all arithmetic to `big.Int`

This second con leads to code that looks like:
```go
func (z *BigInt) And(x, y *BigInt) *BigInt {
	var tmp1, tmp2, tmp3 big.Int
	zi := z.inner(&tmp1)
	zi.And(x.inner(&tmp2), y.inner(&tmp3))
	z.updateInner(zi)
	return z
}
```

This approach compared favorably to the `master` implementation on benchmarks. For instance, the reduction in heap allocations (see PR description) is mostly the same between the two prototype approaches. For the most part, this still led to net speedups across the board compared to `master`.

However, even with the reduction in memory size and on datasets large enough to not fit entirely in my CPU's cache, benchmarks did not present this new approach favorably compared to the approach taken by this PR. It appears that in this performance sensitive code, even with everything inlined, the on-demand inflation and deflation of stack-allocated `big.Int` structs on every operation was too expensive. Here's the comparison between this PR (old) and the approach taken by the [compact representation prototype](https://github.com/nvanbenschoten/apd/commit/654be62f74b7189b16c929bdb561f86a878d71f1#diff-17b95c705c56bfb96739a18a4e13e3d65a8b185f633dbafc27543e55db390cebR213) (new):
```
name                 old time/op    new time/op    delta
GDA/base-10             113µs ± 1%     114µs ± 1%    +0.51%  (p=0.005 n=10+10)
GDA/plus-10            38.3µs ± 0%    41.9µs ± 0%    +9.30%  (p=0.000 n=9+10)
GDA/divideint-10       23.1µs ± 0%    25.7µs ± 1%   +11.03%  (p=0.000 n=10+10)
GDA/remainder-10       41.7µs ± 0%    48.2µs ± 0%   +15.64%  (p=0.000 n=9+9)
GDA/minus-10           8.95µs ± 0%   10.82µs ± 0%   +20.88%  (p=0.000 n=10+10)
GDA/subtract-10         112µs ± 0%     136µs ± 0%   +21.82%  (p=0.000 n=10+9)
GDA/exp-10              130ms ± 0%     158ms ± 0%   +22.03%  (p=0.000 n=10+10)
GDA/add-10              655µs ± 0%     800µs ± 0%   +22.12%  (p=0.000 n=10+10)
GDA/comparetotal-10    24.4µs ± 1%    30.1µs ± 1%   +23.33%  (p=0.000 n=10+10)
GDA/abs-10             7.11µs ± 0%    8.81µs ± 0%   +23.79%  (p=0.000 n=10+10)
GDA/tointegralx-10     22.6µs ± 0%    28.1µs ± 0%   +24.59%  (p=0.000 n=10+7)
GDA/tointegral-10      22.2µs ± 0%    27.7µs ± 0%   +24.75%  (p=0.000 n=10+9)
GDA/multiply-10        56.3µs ± 0%    71.8µs ± 0%   +27.62%  (p=0.000 n=10+8)
GDA/divide-10           377µs ± 0%     486µs ± 0%   +28.87%  (p=0.000 n=10+10)
GDA/randoms-10         2.29ms ± 0%    2.96ms ± 0%   +29.00%  (p=0.000 n=10+9)
GDA/reduce-10          14.2µs ± 0%    18.4µs ± 0%   +29.25%  (p=0.000 n=10+10)
GDA/log10-10            114ms ± 0%     148ms ± 0%   +29.61%  (p=0.000 n=10+9)
GDA/quantize-10        85.0µs ± 0%   110.1µs ± 0%   +29.62%  (p=0.000 n=10+10)
GDA/ln-10              90.1ms ± 0%   117.0ms ± 0%   +29.82%  (p=0.000 n=10+9)
GDA/compare-10         34.5µs ± 1%    44.9µs ± 1%   +30.18%  (p=0.000 n=10+10)
GDA/power-10            233ms ± 0%     304ms ± 0%   +30.39%  (p=0.000 n=9+10)
GDA/squareroot-10      28.7ms ± 0%    37.4ms ± 0%   +30.45%  (p=0.000 n=8+9)
GDA/rounding-10         492µs ± 0%     645µs ± 0%   +31.13%  (p=0.000 n=10+9)
GDA/cuberoot-apd-10    2.14ms ± 0%    2.81ms ± 0%   +31.14%  (p=0.000 n=10+9)
GDA/powersqrt-10        433ms ± 0%     571ms ± 0%   +31.96%  (p=0.000 n=10+10)
```

So I think we'll want to stick with this PR.

yuzefovich (commented):
Wow, impressive work and a very thorough analysis! I haven't looked into the integration into CRDB, but change to `apd` library makes sense to me.

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 11 of 11 files at r4, 8 of 8 files at r5, 1 of 1 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/102#-:-MsbLTKB5t3H8UgomjEj:b-kzsght)* status: all files reviewed, 9 unresolved discussions (waiting on @jordanlewis and @nvanbenschoten)

___
*[bigint.go, line 29 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbGZQ8CBQi2T6OVr2t:-MsbGZQ8CBQi2T6OVr2u:b2k8zm0) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L29)):*
> ```Go
> // The zero value is ready to use.
> // The value must not be copied after first use.
> type BigInt struct {
> ```

What will happen if a new method is added to `big.Int` and we don't add the corresponding method to `BigInt`?
___
*[bigint.go, line 70 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbFyLWAWIzM0aWEFWo:-MsbFyLWAWIzM0aWEFWp:b17y3mf) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L70)):*
> ```Go
> }
> 
> func (b *BigInt) inner() *big.Int {
> ```

nit: a quick comment describing when `inner` vs `innerOrNil` should be used would be helpful.
___
*[bigint.go, line 118 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbEajhAjn4IHH2EuG6:-MsbEajhAjn4IHH2EuG7:b-ejflpw) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L118)):*
> ```Go
> 
> 		// Before doing so, zero out the inline array, in case it had a value
> 		// previously. This is necessary in edge cases where _inner initially
> ```

I didn't get to the tests yet, but this sounds like a good regression test.
___
*[bigint.go, line 193 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbF0NhCH9cREKhRG45:-MsbF0NhCH9cREKhRG46:b6rul4l) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L193)):*
> ```Go
> func (b *BigInt) Bits() []big.Word {
> 	// Don't expose direct access to the big.Int's word slice.
> 	panic("unimplemented")
> ```

nit: maybe also a regression test for this?
___
*[const.go, line 33 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbHgTF8QUHT8sgonVv:-MsbHgTF8QUHT8sgonVw:b7zk6iu) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/const.go#L33)):*
> ```Go
> 	decimalEight     = New(8, 0)
> 
> 	decimalMax = New(math.MaxInt64, 0)
> ```

nit: maybe calling these `decimalMaxInt64` and `decimalMinInt64` would be more descriptive?
___
*[decimal.go, line 362 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbJJjCEqZeMo9O7A2h:-MsbJJjD7twqI6x1lZnK:b-b49u2h) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/decimal.go#L362)):*
> ```Go
> // them with this scaling, along with the scaling. An error can be produced
> // if the resulting scale factor is out of range.
> func upscale(a, b *Decimal, tmp *BigInt) (*BigInt, *BigInt, int32, error) {
> ```

nit: maybe mention what `tmp` should be and what it'll be used for.
___
*[gda_test.go, line 330 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBbeo9FS3oWCsSC5s:-MsbBbeo9FS3oWCsSC5t:b-6rh0td) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L330)):*
> ```Go
> 
> 			b.ResetTimer()
> 			for i := 0; i < b.N; i++ {
> ```

Not your change, but I don't understand the order of these two `for` loops - currently for every benchmark iteration we ran many bench cases, but I would expect the order to be the opposite so that for a particular bench case we'd run as many iterations as needed for that bench case. Am I missing something?
___
*[gda_test.go, line 333 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBt4T9q5-KYR4zXop:-MsbBt4T9q5-KYR4zXoq:buzzown) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L333)):*
> ```Go
> 				for _, bc := range bcs {
> 					// Ignore errors here because the full tests catch them.
> 					var res Decimal
> ```

nit: not sure if it matters, but we could pull out the definition of `res` to be outside of the `for` loops.
___
*[table.go, line 91 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbJlnZCmfUuulQm065:-MsbJlnZCmfUuulQm066:b-8oywje) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/table.go#L91)):*
> ```Go
> 
> 	n := int64(float64(bl) / digitsToBitsRatio)
> 	var tpmE BigInt
> ```

nit: `s/tpm/tmp/`.

yuzefovich:
I spent some time trying to address [the TODO](https://github.com/cockroachdb/cockroach/blob/798a3e544d7ed38c5ea21085054b97cac067f56d/pkg/col/coldata/native_types.go#L109-L110) you mentioned above in order to make the integration easier, and, indeed, it is quite annoying to do.

Another difficulty (apart from decimals becoming the only data type that is operated on by reference) is that in some cases we do the decimal to be stored by value (for example, this is the case of aggregate functions for which we want to inline the decimal into the function itself).

I'm curious to hear @jordanlewis thoughts about this PR, and if we all think it's worthwhile, I'll figure out hot to address that TODO.

nvanbenschoten:
Thanks for taking a look, Yahor. I agree with everything you said. The no-copy limitation appeared to make the integration into the vectorized execution engine difficult for a number of different reasons, mostly stemming from the need to operate on references instead of values. It looked possible, but would likely be a lot of work.

In light of your exploration, I took another look at the "More compact memory representation" alternative. This alternative approach is appealing because it does not use a self-referential pointer, so it can be copied by value safely (though `Decimal.Set` should still be used to avoid aliasing when that is a problem). I mentioned above that the repeat inflation and deflation of stack-allocated `big.Int` structs on every operation was too expensive. We could see this in the benchmark results when compared to this PR.

I explored what it would take to bypass `big.Int` for basic arithmetic on small values, similar to what the `ericlagergren/decimal` library does. As it turns out, the separation between `apd.Decimal` and `apd.BigInt` made this all fairly straightforward — `apd.Decimal` remains entirely oblivious of these optimizations and `apd.BigInt` can continue to fallback to `big.Int` whenever the arithmetic gets hard. Furthermore, it also means that even the complex arithmetic at the `Decimal` level (e.g. `Decimal.Sqrt`) still benefits from fast-paths on simple arithmetic, because these complex `Decimal` operations are composed of many incremental operations on `BigInt`s.

Here's the change to add these fast-paths to the compact memory representation approach: https://github.com/cockroachdb/apd/commit/c5c0544dc082aed7d47c108cc70508a96f46731d. It's not a lot of code, and it benefits from the existing `BigInt` and `Decimal` test coverage.

With that commit, the comparison between the two approaches shift in favor of the compact (and copyable) memory representation (old is this PR, new is https://github.com/cockroachdb/apd/commit/c5c0544dc082aed7d47c108cc70508a96f46731d):
```
name                 old time/op    new time/op    delta
GDA/squareroot-10      28.7ms ± 0%    15.7ms ± 0%   -45.30%  (p=0.000 n=10+10)
GDA/powersqrt-10        433ms ± 0%     240ms ± 0%   -44.45%  (p=0.000 n=10+10)
GDA/rounding-10         493µs ± 0%     297µs ± 0%   -39.68%  (p=0.000 n=10+10)
GDA/divideint-10       23.1µs ± 0%    15.3µs ± 0%   -33.58%  (p=0.000 n=10+10)
GDA/divide-10           377µs ± 0%     260µs ± 0%   -30.98%  (p=0.000 n=8+10)
GDA/remainder-10       41.7µs ± 0%    31.7µs ± 0%   -24.01%  (p=0.000 n=10+10)
GDA/tointegral-10      22.2µs ± 0%    17.7µs ± 0%   -20.29%  (p=0.000 n=10+9)
GDA/randoms-10         2.30ms ± 0%    1.85ms ± 0%   -19.51%  (p=0.000 n=10+10)
GDA/tointegralx-10     22.5µs ± 0%    18.2µs ± 0%   -19.35%  (p=0.000 n=10+10)
GDA/compare-10         34.5µs ± 1%    28.2µs ± 0%   -18.29%  (p=0.000 n=10+10)
GDA/minus-10           8.94µs ± 0%    7.47µs ± 0%   -16.39%  (p=0.000 n=9+9)
GDA/abs-10             7.13µs ± 0%    5.99µs ± 0%   -15.95%  (p=0.000 n=10+10)
GDA/subtract-10         112µs ± 0%      97µs ± 0%   -13.10%  (p=0.000 n=10+10)
GDA/cuberoot-apd-10    2.14ms ± 0%    1.89ms ± 0%   -11.91%  (p=0.000 n=8+10)
GDA/exp-10              130ms ± 0%     118ms ± 1%    -9.34%  (p=0.000 n=8+10)
GDA/ln-10              90.3ms ± 0%    82.3ms ± 0%    -8.79%  (p=0.000 n=10+10)
GDA/log10-10            114ms ± 0%     106ms ± 0%    -7.25%  (p=0.000 n=10+8)
GDA/quantize-10        85.1µs ± 0%    80.1µs ± 0%    -5.89%  (p=0.000 n=10+9)
GDA/power-10            233ms ± 0%     220ms ± 0%    -5.66%  (p=0.000 n=10+9)
GDA/reduce-10          14.3µs ± 0%    13.5µs ± 0%    -5.42%  (p=0.000 n=9+10)
GDA/plus-10            38.4µs ± 0%    37.3µs ± 0%    -2.93%  (p=0.000 n=9+10)
GDA/multiply-10        56.5µs ± 1%    55.8µs ± 1%    -1.19%  (p=0.000 n=10+10)
GDA/comparetotal-10    24.4µs ± 2%    24.4µs ± 2%      ~     (p=0.971 n=10+10)
GDA/base-10             114µs ± 1%     114µs ± 1%    +0.55%  (p=0.019 n=10+10)
GDA/add-10              657µs ± 0%     674µs ± 0%    +2.61%  (p=0.000 n=10+10)
```

_(`GDA/add` is a somewhat unrealistic benchmark because almost all arithmetic takes place on high-precision values that aren't inlined by either approach. `GDA/minus` contains a more realistic distribution)_

I'll look into packaging up that alternative approach and addressing the existing comments of yours that apply to it.

yuzefovich:
Thanks Nathan, this looks very promising! I think I was able to make the switch from the value to the pointer in the vectorized engine (in https://github.com/cockroachdb/cockroach/pull/74469), at least the unit tests seem to work, but I'll be happy if we end up not needing that change :)

nvanbenschoten (commented):
I opened https://github.com/cockroachdb/apd/pull/103 as an alternative to this PR. In that one, I've addressed each of your questions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/102#-:-Msgg_e4DeyBCZgSEhob:b-n4eny2)* status: all files reviewed, 5 unresolved discussions (waiting on @jordanlewis and @yuzefovich)

___
*[bigint.go, line 29 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbGZQ8CBQi2T6OVr2t:-MsgbsgwACiptmD0DrDu:bilm0pn) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L29)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

What will happen if a new method is added to `big.Int` and we don't add the corresponding method to `BigInt`?
</blockquote></details>

(answer the same for either implementation) We're not embedding the `big.Int`, so the methods won't be exported by this type. If we need them, we can add them after the fact.
___
*[bigint.go, line 70 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbFyLWAWIzM0aWEFWo:-MsgcGLE8Rrke9IA31kD:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L70)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: a quick comment describing when `inner` vs `innerOrNil` should be used would be helpful.
</blockquote></details>

Done.
___
*[bigint.go, line 118 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbEajhAjn4IHH2EuG6:-MsgcHn33n8DTadGZF12:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L118)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

I didn't get to the tests yet, but this sounds like a good regression test.
</blockquote></details>

Done.
___
*[bigint.go, line 193 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbF0NhCH9cREKhRG45:-MsgeoZ-3sl4T8-p9KM-:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L193)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: maybe also a regression test for this?
</blockquote></details>

Done.
___
*[const.go, line 33 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbHgTF8QUHT8sgonVv:-Msgca8n3ugwJMwy61fL:bfbq99i) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/const.go#L33)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: maybe calling these `decimalMaxInt64` and `decimalMinInt64` would be more descriptive?
</blockquote></details>

Good point, done.
___
*[decimal.go, line 362 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbJJjCEqZeMo9O7A2h:-MsgctUx6JvGwWxa5rjL:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/decimal.go#L362)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: maybe mention what `tmp` should be and what it'll be used for.
</blockquote></details>

Done.
___
*[gda_test.go, line 330 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBbeo9FS3oWCsSC5s:-Msgd0xM9R7ncpbABo_4:b79f64n) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L330)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

Not your change, but I don't understand the order of these two `for` loops - currently for every benchmark iteration we ran many bench cases, but I would expect the order to be the opposite so that for a particular bench case we'd run as many iterations as needed for that bench case. Am I missing something?
</blockquote></details>

I don't think it makes a difference in practice. Either way, we're running `b.N` iterations of each bench case.
___
*[gda_test.go, line 333 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBt4T9q5-KYR4zXop:-Msgd6WHCRRXx63TlX3c:b-56rrir) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L333)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: not sure if it matters, but we could pull out the definition of `res` to be outside of the `for` loops.
</blockquote></details>

Good point. I pulled this into a vector of `Decimal` inputs and outputs to more realistically model how the vectorized execution engine will evaluate arithmetic. As it turns out, doing so caused the benchmarks to skew even further (by a few percent) in favor of the compact representation, likely because the improved cache locality highlighted the size difference (2 `Decimals` per cache line vs. 1).
___
*[table.go, line 91 at r5](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbJlnZCmfUuulQm065:-Msgdiu25qPRzroBf8nx:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/c4738edeb2c3974f9e2a95fe3dd06c14994c24ef/table.go#L91)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: `s/tpm/tmp/`.
</blockquote></details>

Done.

yuzefovich (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/102)* status: all files reviewed, 2 unresolved discussions (waiting on @jordanlewis and @nvanbenschoten)

___
*[bigint.go, line 29 at r4](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbGZQ8CBQi2T6OVr2t:-Msgi086CK-DgPQTtgOE:b9111hf) ([raw file](https://github.com/cockroachdb/apd/blob/af980d5c68971c10db12919f65054017eb44daa0/bigint.go#L29)):*

(answer the same for either implementation) We're not embedding the `big.Int`, so the methods won't be exported by this type. If we need them, we can add them after the fact.
</blockquote></details>

Indeed, good point.
___
*[gda_test.go, line 330 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBbeo9FS3oWCsSC5s:-Msgi6MUC55cEr2X6-mC:b-jylpyy) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L330)):*

I don't think it makes a difference in practice. Either way, we're running `b.N` iterations of each bench case.
</blockquote></details>

I thought that the value of `b.N` changes for each bench case depending on the variability and length of the benchmark runs with the same name. Is that not the case?

nvanbenschoten (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/102)* status: all files reviewed, 2 unresolved discussions (waiting on @jordanlewis and @yuzefovich)

___
*[gda_test.go, line 330 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBbeo9FS3oWCsSC5s:-MsiM6LCDqVejPz6Bpbw:b-evhe46) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L330)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

I thought that the value of `b.N` changes for each bench case depending on the variability and length of the benchmark runs with the same name. Is that not the case?
</blockquote></details>

Yeah, it's kind of confusing how this works. Each of these "GDA" files contains a series of roughly 1000 operations of varying complexity and we run the entire suite of operations in a given file per benchmark iteration. So at the file level, we have sub-benchmarks (see `b.Run` above), but within a test, we loop over all operations (`benchCase`s) for each benchmark iteration.

I do think this is what we want though because it leads to stable results even with individual operations that vary wildly in cost.

yuzefovich (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/102)* status: all files reviewed, 1 unresolved discussion (waiting on @jordanlewis and @nvanbenschoten)

___
*[gda_test.go, line 330 at r1](https://reviewable.io/reviews/cockroachdb/apd/102#-MsbBbeo9FS3oWCsSC5s:-MskTda24Mc6Hocs0cEh:b-l0z3xc) ([raw file](https://github.com/cockroachdb/apd/blob/06a8216457b2267e8fff4202c9627d3032e18d2c/gda_test.go#L330)):*

Yeah, it's kind of confusing how this works. Each of these "GDA" files contains a series of roughly 1000 operations of varying complexity and we run the entire suite of operations in a given file per benchmark iteration. So at the file level, we have sub-benchmarks (see `b.Run` above), but within a test, we loop over all operations (`benchCase`s) for each benchmark iteration.

I do think this is what we want though because it leads to stable results even with individual operations that vary wildly in cost.
</blockquote></details>

I see, thanks for explaining.

PR #103, opened 2022-01-06
apd: embed small coefficient values in Decimal struct

nvanbenschoten:
Fourth time's the charm.

Replaces cockroachdb/cockroach#74369.
Replaces https://github.com/cockroachdb/apd/pull/101.
Replaces https://github.com/cockroachdb/apd/pull/102.

This is an alternative to #102. At a high-level, it implements the "compact memory representation" described in https://github.com/cockroachdb/apd/pull/102#issuecomment-1005078022. Compared to the approach in that PR, this approach is a) [faster for most operations](https://github.com/cockroachdb/apd/pull/102#issuecomment-1005970829), b) more usable because values can be safely copied, c) half the memory size (32 bytes per `Decimal`, vs. 64).

The memory representation of the Decimal struct in this approach looks like:
```go
type Decimal struct {
    Form     int8
    Negative bool
    Exponent int32
    Coeff    BigInt {
        _inner  *big.Int // nil when value fits in _inline
        _inline [2]uint
    }
} // sizeof = 32
```

With a two-word inline array, any value that would fit in a 128-bit integer (i.e. decimals with a scale-adjusted absolute value up to 2^128 - 1) fit in `_inline`. The indirection through `_inner` is only used for values larger than this.

Before this change, the memory representation of the Decimal struct looked like:
```go
type Decimal struct {
    Form     int64
    Negative bool
    Exponent int32
    Coeff    big.Int {
        neg bool
        abs []big.Word {
            data uintptr ---------------. 
            len  int64                  v
            cap  int64         [uint, uint, ...] // sizeof = variable, but around cap = 4, so 32 bytes
        }
    }
} // sizeof = 48 flat bytes + variable-length heap allocated array
```

----

This commit introduces a performance optimization that embeds small coefficient values directly in their `Decimal` struct, instead of storing these values in a separate heap allocation. It does so by replacing `math/big.Int` with a new wrapper type called `BigInt` that provides an "inline" compact representation optimization.

Each `BigInt` maintains (through `big.Int`) an internal reference to a variable-length integer value, which is represented by a []big.Word. The _inline field and the inner and updateInner methods combine to allow BigInt to inline this variable-length integer array within the BigInt struct when its value is sufficiently small. In the inner method, we point a temporary big.Int's nat slice at this _inline array. big.Int will avoid re-allocating this array until it is provided with a value that exceeds the initial capacity. Later in updateInner, we detect whether the array has been re-allocated. If so, we switch to using the _inner. If not, we continue to use this array.

We set the capacity of the inline array to accommodate any value that would fit in a 128-bit integer (i.e. values with an absolute value up to 2^128 - 1).

This is an alternative to an optimization that many other arbitrary precision decimal libraries have where small coefficient values are stored as numeric fields in their data type's struct. Only when this coefficient value gets large enough do these libraries fall back to a variable-length coefficient with internal indirection. We can see the optimization in practice in the `ericlagergren/decimal` library, where each struct contains a `compact uint64` and an `unscaled big.Int`. Prior concern from the authors of `cockroachdb/apd` regarding this form of compact representation optimization was that it traded performance for complexity. The optimization fractured control flow, leaking out across the library and leading to more complex, error-prone code.

The approach taken in this commit does not have the same issue. All arithmetic on the decimal's coefficient is still deferred to `bit.Int`.

### Impact on benchmarks:

```
name                   old time/op    new time/op    delta
GDA/rounding-10           607µs ± 1%     312µs ± 1%   -48.62%  (p=0.000 n=10+10)
GDA/squareroot-10        31.3ms ± 0%    16.4ms ± 1%   -47.47%  (p=0.000 n=7+10)
GDA/powersqrt-10          445ms ± 0%     252ms ± 1%   -43.39%  (p=0.000 n=9+10)
GDA/tointegral-10        31.6µs ± 1%    18.0µs ± 1%   -43.01%  (p=0.000 n=9+10)
GDA/tointegralx-10       31.9µs ± 1%    18.4µs ± 1%   -42.31%  (p=0.000 n=8+9)
GDA/remainder-10         55.4µs ± 1%    32.8µs ± 1%   -40.79%  (p=0.000 n=10+10)
GDA/abs-10               10.1µs ± 1%     6.2µs ± 1%   -38.95%  (p=0.000 n=10+9)
GDA/subtract-10           150µs ± 1%      97µs ± 1%   -35.72%  (p=0.000 n=10+10)
GDA/divideint-10         24.2µs ± 0%    15.9µs ± 1%   -34.24%  (p=0.000 n=9+10)
GDA/randoms-10           2.92ms ± 1%    1.92ms ± 1%   -34.21%  (p=0.000 n=10+10)
GDA/minus-10             11.2µs ± 1%     7.7µs ± 1%   -31.10%  (p=0.000 n=9+10)
GDA/quantize-10           113µs ± 0%      81µs ± 1%   -28.01%  (p=0.000 n=10+10)
GDA/compare-10           40.5µs ± 1%    29.4µs ± 1%   -27.46%  (p=0.000 n=10+9)
GDA/divide-10             365µs ± 1%     272µs ± 1%   -25.47%  (p=0.000 n=9+9)
GDA/reduce-10            17.8µs ± 1%    14.0µs ± 1%   -21.18%  (p=0.000 n=10+10)
GDA/add-10                763µs ± 1%     610µs ± 1%   -19.98%  (p=0.000 n=10+9)
GDA/multiply-10          68.5µs ± 1%    57.0µs ± 1%   -16.84%  (p=0.000 n=9+10)
GDA/comparetotal-10      29.4µs ± 1%    25.3µs ± 2%   -13.82%  (p=0.000 n=9+10)
GDA/plus-10              42.0µs ± 1%    36.8µs ± 1%   -12.33%  (p=0.000 n=9+10)
GDA/cuberoot-apd-10      2.06ms ± 3%    1.99ms ± 1%    -3.75%  (p=0.000 n=10+9)
GDA/exp-10                128ms ± 0%     126ms ± 0%    -1.01%  (p=0.000 n=8+10)
GDA/ln-10                83.9ms ± 0%    86.7ms ± 1%    +3.36%  (p=0.000 n=8+10)
GDA/log10-10              106ms ± 1%     112ms ± 1%    +5.51%  (p=0.000 n=9+10)
GDA/base-10               110µs ± 0%     116µs ± 1%    +5.76%  (p=0.000 n=8+9)
GDA/power-10              218ms ± 0%     231ms ± 1%    +6.00%  (p=0.000 n=9+10)

name                   old alloc/op   new alloc/op   delta
GDA/abs-10               2.33kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/compare-10           10.1kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)
GDA/comparetotal-10      7.11kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/minus-10             2.43kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/reduce-10            2.22kB ± 0%    0.00kB       -100.00%  (p=0.000 n=10+10)
GDA/remainder-10         22.4kB ± 0%     0.1kB ± 0%   -99.57%  (p=0.000 n=10+10)
GDA/rounding-10           246kB ± 0%       3kB ± 0%   -98.75%  (p=0.000 n=10+10)
GDA/divideint-10         6.51kB ± 0%    0.10kB ± 0%   -98.53%  (p=0.000 n=10+10)
GDA/squareroot-10        8.30MB ± 0%    0.31MB ± 0%   -96.28%  (p=0.000 n=10+10)
GDA/randoms-10           1.16MB ± 0%    0.05MB ± 0%   -95.82%  (p=0.000 n=10+10)
GDA/powersqrt-10         77.6MB ± 0%     6.0MB ± 0%   -92.26%  (p=0.000 n=10+10)
GDA/divide-10            74.3kB ± 0%     9.7kB ± 0%   -86.95%  (p=0.000 n=10+10)
GDA/quantize-10          42.3kB ± 0%     9.2kB ± 0%   -78.23%  (p=0.002 n=8+10)
GDA/tointegralx-10       19.2kB ± 0%     6.6kB ± 0%   -65.78%  (p=0.000 n=10+10)
GDA/tointegral-10        19.1kB ± 0%     6.6kB ± 0%   -65.61%  (p=0.000 n=10+10)
GDA/subtract-10           100kB ± 0%      35kB ± 0%   -64.64%  (p=0.000 n=10+10)
GDA/multiply-10          35.5kB ± 0%    15.7kB ± 0%   -55.83%  (p=0.000 n=10+10)
GDA/cuberoot-apd-10       261kB ± 0%     129kB ± 0%   -50.46%  (p=0.000 n=9+10)
GDA/add-10                712kB ± 0%     459kB ± 0%   -35.53%  (p=0.000 n=9+9)
GDA/ln-10                10.2MB ± 0%     7.2MB ± 0%   -29.57%  (p=0.000 n=10+10)
GDA/log10-10             12.5MB ± 0%     9.1MB ± 0%   -27.14%  (p=0.000 n=10+10)
GDA/power-10             27.0MB ± 0%    20.7MB ± 0%   -23.35%  (p=0.000 n=9+10)
GDA/plus-10              46.6kB ± 0%    41.0kB ± 0%   -11.95%  (p=0.000 n=10+10)
GDA/exp-10               61.3MB ± 0%    60.6MB ± 0%    -1.09%  (p=0.000 n=10+10)
GDA/base-10              24.4kB ± 0%    26.2kB ± 0%    +7.35%  (p=0.000 n=10+10)

name                   old allocs/op  new allocs/op  delta
GDA/abs-10                  151 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/compare-10              638 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/comparetotal-10         254 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/minus-10                164 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/reduce-10               187 ± 0%         0       -100.00%  (p=0.000 n=10+10)
GDA/remainder-10            923 ± 0%         2 ± 0%   -99.78%  (p=0.000 n=10+10)
GDA/divideint-10            357 ± 0%         2 ± 0%   -99.44%  (p=0.000 n=10+10)
GDA/rounding-10           9.20k ± 0%     0.10k ± 0%   -98.96%  (p=0.000 n=10+10)
GDA/randoms-10            40.4k ± 0%      0.7k ± 0%   -98.26%  (p=0.000 n=10+10)
GDA/squareroot-10          275k ± 0%        6k ± 0%   -97.88%  (p=0.000 n=10+10)
GDA/quantize-10           1.70k ± 0%     0.05k ± 0%   -96.89%  (p=0.000 n=10+10)
GDA/tointegralx-10          677 ± 0%        48 ± 0%   -92.91%  (p=0.000 n=10+10)
GDA/tointegral-10           665 ± 0%        48 ± 0%   -92.78%  (p=0.000 n=10+10)
GDA/subtract-10           2.85k ± 0%     0.21k ± 0%   -92.63%  (p=0.000 n=10+10)
GDA/powersqrt-10          2.63M ± 0%     0.20M ± 0%   -92.20%  (p=0.000 n=10+10)
GDA/divide-10             2.51k ± 0%     0.20k ± 0%   -91.94%  (p=0.000 n=10+10)
GDA/add-10                13.2k ± 0%      4.2k ± 0%   -68.01%  (p=0.000 n=10+10)
GDA/multiply-10             931 ± 0%       312 ± 0%   -66.49%  (p=0.000 n=10+10)
GDA/cuberoot-apd-10       7.67k ± 0%     2.62k ± 0%   -65.91%  (p=0.000 n=10+10)
GDA/plus-10                 587 ± 0%       260 ± 0%   -55.71%  (p=0.000 n=10+10)
GDA/ln-10                  269k ± 0%      146k ± 0%   -45.71%  (p=0.000 n=10+10)
GDA/log10-10               326k ± 0%      186k ± 0%   -42.98%  (p=0.000 n=10+10)
GDA/power-10               699k ± 0%      423k ± 0%   -39.52%  (p=0.000 n=10+10)
GDA/exp-10                 911k ± 0%      696k ± 0%   -23.62%  (p=0.000 n=10+10)
GDA/base-10               2.11k ± 0%     2.52k ± 0%   +19.14%  (p=0.000 n=10+10)
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -15,8 +15,6 @@ jobs:
     strategy:
       matrix:
         go:
-          - '1.11'
-          - '1.12'
           - '1.13'
           - '1.14'
           - '1.15'
@@ -38,11 +36,12 @@ jobs:
         run: go test -v ./...
 
       - name: 'Vet'
-        run: go vet ./...
+        # -unsafeptr=false is needed because of the noescape function in bigint.go.
+        run: go vet -unsafeptr=false ./...
 
       - name: 'Staticcheck'
         # staticcheck requires go1.14.
-        if: ${{ matrix.go != '1.11' && matrix.go != '1.12' && matrix.go != '1.13' }}
+        if: ${{ matrix.go != '1.13' }}
         run: |
           go get honnef.co/go/tools/cmd/staticcheck
           staticcheck ./...
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ apd is an arbitrary-precision decimal package for Go.
 - **Good performance**. Operations will either be fast enough or will produce an error if they will be slow. This prevents edge-case operations from consuming lots of CPU or memory.
 - **Condition flags and traps**. All operations will report whether their result is exact, is rounded, is over- or under-flowed, is [subnormal](https://en.wikipedia.org/wiki/Denormal_number), or is some other condition. `apd` supports traps which will trigger an error on any of these conditions. This makes it possible to guarantee exactness in computations, if needed.
 
-`apd` has three main types. The first is [`BigInt`](https://godoc.org/github.com/cockroachdb/apd#BigInt) which is a wrapper around `big.Int` that exposes an identical API while reducing memory allocations. `BigInt` does so by using an inline array to back the `big.Int`'s variable-length value slice when the integer's absolute value is sufficiently small. 
+`apd` has three main types.
+
+The first is [`BigInt`](https://godoc.org/github.com/cockroachdb/apd#BigInt) which is a wrapper around `big.Int` that exposes an identical API while reducing memory allocations. `BigInt` does so by using an inline array to back the `big.Int`'s variable-length value when the integer's absolute value is sufficiently small. `BigInt` also contains fast-paths that allow it to perform basic arithmetic directly on this inline array, only falling back to `big.Int` when the arithmetic gets complex or takes place on large values.
 
 The second is [`Decimal`](https://godoc.org/github.com/cockroachdb/apd#Decimal) which holds the values of decimals. It is simple and uses a `BigInt` with an exponent to describe values. Most operations on `Decimal`s can’t produce errors as they work directly on the underlying `big.Int`. Notably, however, there are no arithmetic operations on `Decimal`s.
 
--- a/README.md
+++ b/README.md
@@ -12,13 +12,15 @@ apd is an arbitrary-precision decimal package for Go.
 - **Good performance**. Operations will either be fast enough or will produce an error if they will be slow. This prevents edge-case operations from consuming lots of CPU or memory.
 - **Condition flags and traps**. All operations will report whether their result is exact, is rounded, is over- or under-flowed, is [subnormal](https://en.wikipedia.org/wiki/Denormal_number), or is some other condition. `apd` supports traps which will trigger an error on any of these conditions. This makes it possible to guarantee exactness in computations, if needed.
 
-`apd` has two main types. The first is [`Decimal`](https://godoc.org/github.com/cockroachdb/apd#Decimal) which holds the values of decimals. It is simple and uses a `big.Int` with an exponent to describe values. Most operations on `Decimal`s can’t produce errors as they work directly on the underlying `big.Int`. Notably, however, there are no arithmetic operations on `Decimal`s.
+`apd` has three main types. The first is [`BigInt`](https://godoc.org/github.com/cockroachdb/apd#BigInt) which is a wrapper around `big.Int` that exposes an identical API while reducing memory allocations. `BigInt` does so by using an inline array to back the `big.Int`'s variable-length value slice when the integer's absolute value is sufficiently small. 
 
-The second main type is [`Context`](https://godoc.org/github.com/cockroachdb/apd#Context), which is where all arithmetic operations are defined. A `Context` describes the precision, range, and some other restrictions during operations. These operations can all produce failures, and so return errors.
+The second is [`Decimal`](https://godoc.org/github.com/cockroachdb/apd#Decimal) which holds the values of decimals. It is simple and uses a `BigInt` with an exponent to describe values. Most operations on `Decimal`s can’t produce errors as they work directly on the underlying `big.Int`. Notably, however, there are no arithmetic operations on `Decimal`s.
+
+The third main type is [`Context`](https://godoc.org/github.com/cockroachdb/apd#Context), which is where all arithmetic operations are defined. A `Context` describes the precision, range, and some other restrictions during operations. These operations can all produce failures, and so return errors.
 
 `Context` operations, in addition to errors, return a [`Condition`](https://godoc.org/github.com/cockroachdb/apd#Condition), which is a bitfield of flags that occurred during an operation. These include overflow, underflow, inexact, rounded, and others. The `Traps` field of a `Context` can be set which will produce an error if the corresponding flag occurs. An example of this is given below.
 
 See the [examples](https://godoc.org/github.com/cockroachdb/apd#pkg-examples) for some operations that were previously difficult to perform in Go.
 
 ## Documentation
-https://pkg.go.dev/github.com/cockroachdb/apd/v2?tab=doc
+https://pkg.go.dev/github.com/cockroachdb/apd/v3?tab=doc
--- a/bigint.go
+++ b/bigint.go
@@ -294,6 +294,9 @@ func mulInline(xVal, yVal uint, xNeg, yNeg bool) (zVal uint, zNeg, ok bool) {
 
 //gcassert:inline
 func quoInline(xVal, yVal uint, xNeg, yNeg bool) (quoVal uint, quoNeg, ok bool) {
+	if yVal == 0 { // divide by 0
+		return 0, false, false
+	}
 	quo := xVal / yVal
 	neg := xNeg != yNeg
 	return quo, neg, true
@@ -301,6 +304,9 @@ func quoInline(xVal, yVal uint, xNeg, yNeg bool) (quoVal uint, quoNeg, ok bool) 
 
 //gcassert:inline
 func remInline(xVal, yVal uint, xNeg, yNeg bool) (remVal uint, remNeg, ok bool) {
+	if yVal == 0 { // divide by 0
+		return 0, false, false
+	}
 	rem := xVal % yVal
 	return rem, xNeg, true
 }
@@ -361,7 +367,7 @@ func (z *BigInt) AndNot(x, y *BigInt) *BigInt {
 // Append calls (big.Int).Append.
 func (z *BigInt) Append(buf []byte, base int) []byte {
 	var tmp1 big.Int
-	return z.inner(&tmp1).Append(buf, base)
+	return z.innerOrNil(&tmp1).Append(buf, base)
 }
 
 // Binomial calls (big.Int).Binomial.
@@ -770,10 +776,10 @@ func (z *BigInt) Set(x *BigInt) *BigInt {
 }
 
 // SetBit calls (big.Int).SetBit.
-func (z *BigInt) SetBit(x *BigInt, i int, v uint) *BigInt {
+func (z *BigInt) SetBit(x *BigInt, i int, b uint) *BigInt {
 	var tmp1, tmp2 big.Int
 	zi := z.inner(&tmp1)
-	zi.SetBit(x.inner(&tmp2), i, v)
+	zi.SetBit(x.inner(&tmp2), i, b)
 	z.updateInner(zi)
 	return z
 }
@@ -863,7 +869,7 @@ func (z *BigInt) Sqrt(x *BigInt) *BigInt {
 // String calls (big.Int).String.
 func (z *BigInt) String() string {
 	var tmp1 big.Int
-	return z.inner(&tmp1).String()
+	return z.innerOrNil(&tmp1).String()
 }
 
 // Sub calls (big.Int).Sub.
@@ -886,7 +892,7 @@ func (z *BigInt) Sub(x, y *BigInt) *BigInt {
 // Text calls (big.Int).Text.
 func (z *BigInt) Text(base int) string {
 	var tmp1 big.Int
-	return z.inner(&tmp1).Text(base)
+	return z.innerOrNil(&tmp1).Text(base)
 }
 
 // TrailingZeroBits calls (big.Int).TrailingZeroBits.
--- a/bigint.go
+++ b/bigint.go
@@ -1,0 +1,688 @@
+// Copyright 2022 The Cockroach Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package apd
+
+import (
+	"fmt"
+	"math/big"
+	"math/rand"
+	"unsafe"
+)
+
+// The inlineWords capacity is set to accommodate any value that would fit in a
+// 128-bit integer (i.e. values with an absolute value up to 2^128 - 1).
+const inlineWords = 2
+
+// BigInt is a wrapper around big.Int. It minimizes memory allocation by using
+// an inline array to back the big.Int's variable-length "nat" slice when the
+// integer's value is sufficiently small.
+// The zero value is ready to use.
+type BigInt struct {
+	// A wrapped big.Int. Only set to the BigInt's value when the value exceeds
+	// what is representable in the _inline array.
+	//
+	// When the BigInt's value is still small enough to use the _inline array,
+	// this field doubles as integer's negative flag. See negSentinel.
+	//
+	// Methods should access this field through inner.
+	_inner *big.Int
+
+	// The inlined backing array use for short-lived, stack-allocated big.Int
+	// structs during arithmetic when the value is small.
+	//
+	// Each BigInt maintains (through big.Int) an internal reference to a
+	// variable-length integer value, which is represented by a []big.Word. The
+	// _inline field and the inner and updateInner methods combine to allow
+	// BigInt to inline this variable-length integer array within the BigInt
+	// struct when its value is sufficiently small. In the inner method, we
+	// point a temporary big.Int's nat slice at this _inline array. big.Int will
+	// avoid re-allocating this array until it is provided with a value that
+	// exceeds the initial capacity. Later in updateInner, we detect whether the
+	// array has been re-allocated. If so, we switch to using the _inner. If
+	// not, we continue to use this array.
+	_inline [inlineWords]big.Word
+}
+
+// NewBigInt allocates and returns a new BigInt set to x.
+//
+// NOTE: BigInt jumps through hoops to avoid escaping to the heap. As such, most
+// users of BigInt should not need this function. They should instead declare a
+// zero-valued BigInt directly on the stack and interact with references to this
+// stack-allocated value. Recall that the zero-valued BigInt is ready to use.
+func NewBigInt(x int64) *BigInt {
+	return new(BigInt).SetInt64(x)
+}
+
+// Set as the value of BigInt._inner as a "sentinel" flag to indicate that a
+// BigInt is negative ((big.Int).Sign() < 0) but the absolute value is still
+// small enough to represent in the _inline array.
+var negSentinel = new(big.Int)
+
+// The memory representation of big.Int. Used for unsafe modification below.
+type intStruct struct {
+	neg bool
+	abs []big.Word
+}
+
+// noescape hides a pointer from escape analysis. noescape is the identity
+// function but escape analysis doesn't think the output depends on the input.
+// noescape is inlined and currently compiles down to zero instructions.
+//
+// USE CAREFULLY!
+//
+// This was copied from strings.Builder, which has identical code which was
+// itself copied from the runtime.
+// For more, see issues #23382 and #7921 in github.com/golang/go.
+//go:nosplit
+//go:nocheckptr
+func noescape(p unsafe.Pointer) unsafe.Pointer {
+	x := uintptr(p)
+	//lint:ignore SA4016 intentional no-op to hide pointer from escape analysis.
+	return unsafe.Pointer(x ^ 0)
+}
+
+// inner returns the BigInt's current value as a *big.Int.
+//
+// NOTE: this was carefully written to permit function inlining. Modify with
+// care.
+func (z *BigInt) inner(tmp *big.Int) *big.Int {
+	// Point the big.Int at the inline array. When doing so, use noescape to
+	// avoid forcing the BigInt to escape to the heap. Go's escape analysis
+	// struggles with self-referential pointers, and it can't prove that we
+	// only assign _inner to a heap-allocated object (which must not contain
+	// pointers that reference the stack or the GC explodes) if the big.Int's
+	// backing array has been re-allocated onto the heap first.
+	//
+	// NOTE: SetBits sets the neg field to false, so this must come before the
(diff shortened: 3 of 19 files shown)


yuzefovich (commented):
This is awesome! Kudos for figuring all this complexity out. I only have a handful of questions / comments.

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 13 of 13 files at r4, 2 of 2 files at r5, 8 of 8 files at r6, 1 of 1 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/103#-:-Msgx2yk1wTFPHO5Kdh7:brfwpj1)* status: all files reviewed, 6 unresolved discussions (waiting on @jordanlewis and @nvanbenschoten)

___
*[bigint.go, line 97 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgwGhg1U3nqczBIBVR:-MsgwGhg1U3nqczBIBVS:bynk2l0) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L97)):*
> ```Go
> // inner returns the BigInt's current value as a *big.Int.
> //
> // NOTE: this was carefully written to permit function inlining. Modify with
> ```

I think you're probably aware of `gcassert` library that Jordan wrote, and it seems like it could be useful in this file.
___
*[bigint.go, line 107 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-Msgm7zZ1us_QDflQx7Z:-Msgm7zZ1us_QDflQx7_:bv2tex9) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L107)):*
> ```Go
> 	// backing array has been re-allocated onto the heap first.
> 	//
> 	// NOTE: SetBits set the neg field to false, so this must come before the
> ```

nit: `s/SetBits set/SetBits sets/`.
___
*[bigint.go, line 121 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgmeTw7TWOcB_XWVhK:-MsgmeTw7TWOcB_XWVhL:b-uw38nb) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L121)):*
> ```Go
> 		// use unsafe because (*big.Int).Neg is too complex and prevents this
> 		// method from being inlined.
> 		(*intStruct)(unsafe.Pointer(tmp)).neg = true
> ```

Is it worth adding a regression test for this? Like I imagine defining non-inlinable `innerSlow` function that doesn't use this unsafe stuff, and then comparing the results of `inner` and `innerSlow`?
___
*[bigint.go, line 252 at r5](https://reviewable.io/reviews/cockroachdb/apd/103#-Msgr9iLBSGMRLQn0Ty2:-Msgr9iLBSGMRLQn0Ty3:b-nfj0uj) ([raw file](https://github.com/cockroachdb/apd/blob/7edae273c7d5c73ca5420376a1553b4a9fb3e2ed/bigint.go#L252)):*
> ```Go
> 	if xNeg == yNeg {
> 		sum, carry := bits.Add(xVal, yVal, 0)
> 		if carry != 0 { // overflow
> ```

Do you think it would be faster to remove the `if` branch and rely on the always computed `ok` value? I.e. something like:
```
if xNeg == yNed {
  sum, carry := bits.Add(xVal, yVal, 0)
  return sum, xNeg, carry == 0
}
...
```
___
*[gda_test.go, line 332 at r1](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgiwRMA55xK2SbMZI7:-MsgiwRMA55xK2SbMZI8:buicioj) ([raw file](https://github.com/cockroachdb/apd/blob/7b4be850fe0c6352fc047a33192c024a4e364f15/gda_test.go#L332)):*
> ```Go
> 			op1s := make([]Decimal, len(bcs))
> 			op2s := make([]Decimal, len(bcs))
> 			res := make([]Decimal, b.N*len(bcs))
> ```

I'm still a bit confused about this allocation - is that not sufficient to have a single global `Decimal` that we pass into `Run`? IIUC this `Decimal` is used to write the result of operation into, and we don't have to reset it, and it looks like we don't run these benchmarks in parallel either.
___
*[README.md, line 15 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgpRmE6K6YBFhk0ivO:-MsgpRmF9UgwXwnmlv7g:b-pbfzwt) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/README.md#L15)):*
> ```Markdown
> - **Condition flags and traps**. All operations will report whether their result is exact, is rounded, is over- or under-flowed, is [subnormal](https://en.wikipedia.org/wiki/Denormal_number), or is some other condition. `apd` supports traps which will trigger an error on any of these conditions. This makes it possible to guarantee exactness in computations, if needed.
> 
> `apd` has three main types. The first is [`BigInt`](https://godoc.org/github.com/cockroachdb/apd#BigInt) which is a wrapper around `big.Int` that exposes an identical API while reducing memory allocations. `BigInt` does so by using an inline array to back the `big.Int`'s variable-length value slice when the integer's absolute value is sufficiently small. 
> ```

super nit: "variable-length value slice" sounds a bit off.

nit: dangling spaces at the end.

nvanbenschoten (commented):
TFTR!

I had a much easier time hooking this into Cockroach now that these values can be copied around freely. For the most part, it was a clear drop-in replacement.

I was thinking about testing out the change on a few benchmarks. Do you think TPC-DS would be a good one to test it on? That workload seems to perform a large number of aggregations, many of which are on decimal columns. Do you happen to know the individual queries well enough to have suggestions for which ones to look at first?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/103#-:-MsiSbCOAVVb-WQJapFq:b-p8g3t6)* status: 2 of 15 files reviewed, 5 unresolved discussions (waiting on @jordanlewis and @yuzefovich)

___
*[bigint.go, line 97 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgwGhg1U3nqczBIBVR:-MsiOtRm2A4J6SH-5_8g:blx5b3h) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L97)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

I think you're probably aware of `gcassert` library that Jordan wrote, and it seems like it could be useful in this file.
</blockquote></details>

Good point! I added them in.
___
*[bigint.go, line 107 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-Msgm7zZ1us_QDflQx7Z:-MsiOMIC2Z73hAObvR5Q:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L107)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: `s/SetBits set/SetBits sets/`.
</blockquote></details>

Done.
___
*[bigint.go, line 121 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgmeTw7TWOcB_XWVhK:-MsiONac9pjX25vQRoHx:b-t6eqym) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L121)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

Is it worth adding a regression test for this? Like I imagine defining non-inlinable `innerSlow` function that doesn't use this unsafe stuff, and then comparing the results of `inner` and `innerSlow`?
</blockquote></details>

Without this line, most of the library falls over, but I completely agree with your point. I took a slightly different approach though. Instead of comparing against a slow path, I wrote a new test using Go's property-based testing framework to assert that all of the methods on `apd.BigInt` and `math/big.Int` have identical behavior for all inputs. This is in a new commit that I added on to the end of this PR (which is getting a bit big — sorry about that).
___
*[bigint.go, line 252 at r5](https://reviewable.io/reviews/cockroachdb/apd/103#-Msgr9iLBSGMRLQn0Ty2:-MsiQp0H5382sReaQY1M:b1plf3f) ([raw file](https://github.com/cockroachdb/apd/blob/7edae273c7d5c73ca5420376a1553b4a9fb3e2ed/bigint.go#L252)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

Do you think it would be faster to remove the `if` branch and rely on the always computed `ok` value? I.e. something like:
```
if xNeg == yNed {
  sum, carry := bits.Add(xVal, yVal, 0)
  return sum, xNeg, carry == 0
}
...
```
</blockquote></details>

Nice, I think so. Done here and in the inline multiplication arithmetic.
___
*[gda_test.go, line 332 at r1](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgiwRMA55xK2SbMZI7:-MsiOx-dB6UY9cuCegOo:bwhqmxz) ([raw file](https://github.com/cockroachdb/apd/blob/7b4be850fe0c6352fc047a33192c024a4e364f15/gda_test.go#L332)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

I'm still a bit confused about this allocation - is that not sufficient to have a single global `Decimal` that we pass into `Run`? IIUC this `Decimal` is used to write the result of operation into, and we don't have to reset it, and it looks like we don't run these benchmarks in parallel either.
</blockquote></details>

Yes, we could have a single global `Decimal` that we pass into `Run`. The reason I wrote it like this was so that it would more closely resemble real-world use. A few of these different approaches were deferring work to the first time a decimal was touched (allocating on the heap, lazily hooking up a self-referential pointer, etc). Re-using the same result would have amortized this cost when doing so isn't appropriate for what we want to measure here.
___
*[README.md, line 15 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgpRmE6K6YBFhk0ivO:-MsiPe01EzHP4HV0kft4:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/README.md#L15)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

super nit: "variable-length value slice" sounds a bit off.

nit: dangling spaces at the end.
</blockquote></details>

Done.

yuzefovich (approved):
Awesome work! :lgtm:

In terms of benchmarking, it does look like TPC-DS would be a good option, but unfortunately I never analyzed the individual queries to be able to recommend a subset of them to run here. I have been periodically running all of the queries (like once in six months) and updating the expected runtimes in https://github.com/cockroachdb/cockroach/issues/37464#issue-442940968. The information there hopefully can give you some starting point.

> Reviewed 13 of 13 files at r9, 1 of 1 files at r10, 2 of 2 files at r11, 8 of 8 files at r12, 1 of 1 files at r13, 4 of 4 files at r14, 3 of 3 files at r15, 2 of 2 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/103#-:-Msk_PXL0toG03y7A2kN:b9pncv)* status: all files reviewed, 4 unresolved discussions (waiting on @jordanlewis and @nvanbenschoten)

___
*[bigint.go, line 97 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgwGhg1U3nqczBIBVR:-MskUrC_1pEmMIl3kogJ:b-blxs55) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L97)):*

Good point! I added them in.
</blockquote></details>

There is a bit more work needed - we need to pull in `gcassert` as a dependency for testing and then write a test that actually asserts things (e.g. this is what we do in CRDB https://github.com/cockroachdb/cockroach/blob/5c27c74d592782e5911f4e85a6c2e56c2271fe5a/pkg/testutils/lint/lint_test.go#L1972-L2000).
___
*[bigint_test.go, line 545 at r16](https://reviewable.io/reviews/cockroachdb/apd/103#-MskYsivAsXEBy4JOdSJ:-MskYsivAsXEBy4JOdSK:b-4ms3s) ([raw file](https://github.com/cockroachdb/apd/blob/57503b9d7ca92d4376fccc6899d3760c75e930a1/bigint_test.go#L545)):*
> ```Go
> 
> 	t.Run("Scan", func(t *testing.T) {
> 		// Call indirectly through fmt.Scan.
> ```

nit: `fmt.Scan` here and `fmt.Sscan` below do not align.
___
*[bigint_test.go, line 771 at r16](https://reviewable.io/reviews/cockroachdb/apd/103#-MskZLos3mNcvSR16PkN:-MskZLos3mNcvSR16PkO:bcymi4a) ([raw file](https://github.com/cockroachdb/apd/blob/57503b9d7ca92d4376fccc6899d3760c75e930a1/bigint_test.go#L771)):*
> ```Go
> }
> 
> // number is a quick.Generator for decimal numbers.
> ```

nit: I think we're generating integer numbers, not decimal, right?
___
*[gda_test.go, line 332 at r1](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgiwRMA55xK2SbMZI7:-MskVev9A9Crzc0mNzpT:bg2j5hf) ([raw file](https://github.com/cockroachdb/apd/blob/7b4be850fe0c6352fc047a33192c024a4e364f15/gda_test.go#L332)):*

Yes, we could have a single global `Decimal` that we pass into `Run`. The reason I wrote it like this was so that it would more closely resemble real-world use. A few of these different approaches were deferring work to the first time a decimal was touched (allocating on the heap, lazily hooking up a self-referential pointer, etc). Re-using the same result would have amortized this cost when doing so isn't appropriate for what we want to measure here.
</blockquote></details>

Makes sense, thanks.

nvanbenschoten (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/103)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @jordanlewis and @yuzefovich)

___
*[bigint.go, line 97 at r4](https://reviewable.io/reviews/cockroachdb/apd/103#-MsgwGhg1U3nqczBIBVR:-Mslcebx1CtTNGLvLCt9:b-ioql29) ([raw file](https://github.com/cockroachdb/apd/blob/cfc8cc45b6421c57a02b011c16833ff8d5bc3a38/bigint.go#L97)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

There is a bit more work needed - we need to pull in `gcassert` as a dependency for testing and then write a test that actually asserts things (e.g. this is what we do in CRDB https://github.com/cockroachdb/cockroach/blob/5c27c74d592782e5911f4e85a6c2e56c2271fe5a/pkg/testutils/lint/lint_test.go#L1972-L2000).
</blockquote></details>

I was hoping to be able to add this to CI in https://github.com/cockroachdb/apd/pull/106, but for now, it looks like it's blocked on https://github.com/jordanlewis/gcassert/pull/3. I'd like to use gcassert in more places in this library, so I'll keep pulling on that thread.
___
*[bigint_test.go, line 545 at r16](https://reviewable.io/reviews/cockroachdb/apd/103#-MskYsivAsXEBy4JOdSJ:-MsldKUyEYv5oeztJp9y:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/57503b9d7ca92d4376fccc6899d3760c75e930a1/bigint_test.go#L545)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: `fmt.Scan` here and `fmt.Sscan` below do not align.
</blockquote></details>

Done.
___
*[bigint_test.go, line 771 at r16](https://reviewable.io/reviews/cockroachdb/apd/103#-MskZLos3mNcvSR16PkN:-MsldQAg6T8UO-R2xyGR:b-896fix) ([raw file](https://github.com/cockroachdb/apd/blob/57503b9d7ca92d4376fccc6899d3760c75e930a1/bigint_test.go#L771)):*
<details><summary><i>Previously, yuzefovich (Yahor Yuzefovich) wrote…</i></summary><blockquote>

nit: I think we're generating integer numbers, not decimal, right?
</blockquote></details>

Done.

PR #109, opened 2022-01-08
apd: add inline fast-path to BigInt.Format

nvanbenschoten:
This commit adds an inline fast-path to (*BigInt).Format. This method is used in CockroachDB's `encoding.decimalEandM` function, so it's worth optimizing.

Without the fast-path, `(*big.Int).Format` performs at least one [heap allocation to store a temp value](https://github.com/golang/go/blob/7f3eb6182ba72b1b45cf313a01b9c7d4b374255f/src/math/big/natconv.go#L331-L332), which shows up in heap profiles:

Screen Shot 2022-01-07 at 6 59 19 PM

The commit also removes a few calls to `(*BigInt).innerOrNil` in performance-sensitive methods, because this function prevents `(*BigInt).inner` from being inlined.

### DIFF
--- a/bigint.go
+++ b/bigint.go
@@ -19,6 +19,7 @@ import (
 	"math/big"
 	"math/bits"
 	"math/rand"
+	"strconv"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


yuzefovich (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/109#-:-MsrAJmfCCDuy0K8aqUy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nvanbenschoten)

PR #110, opened 2022-01-09
apd: support and test 32-bit architectures and ARM architectures

nvanbenschoten:
This PR switches the `BigInt` fast-paths from operating on uint values to uint64 values. This allows them to work on 32-bit architectures for any value that can fit in a 64-bit unsigned integer.

There are two benefits of this. The first is that it unifies the arithmetic fast-paths across the architectures so that they perform more similarly. The second is that it removes some dead code on 64-bit architectures so that we can avoid issues with jordanlewis/gcassert#3.

The PR then updates CI to test on 32-bit and 64-bit ARM. I'd like to also test on 32-bit x86, but this is blocked on https://github.com/uraimo/run-on-arch-action/issues/57.


yuzefovich (approved):
:lgtm: although it's the first time I'm looking at the CI changes, so I'm more or less trusting you that it behaves as expected :)

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/110#-:-Mt4FJJbByG5gmTE0Prs:b-dq6vk1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nvanbenschoten)

nvanbenschoten:
TFTR!

PR #111, opened 2022-01-09
apd: use gcassert to check that tmp big.Ints don't escape, add to CI

nvanbenschoten:
This commit adds `gcassert:noescape` directives to `bigint.go` to verify
that the temporary big.Int values that are intended to be stack-allocated
do not escape to the heap.

The commit then adds `gcassert` to CI.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -65,6 +65,14 @@ jobs:
           go get honnef.co/go/tools/cmd/staticcheck
           staticcheck ./...
 
+      - name: 'GCAssert'
+        # Only run gcassert on the latest versions of Go. Inlining heuristics
+        # change from version to version.
+        if: ${{ matrix.arch == 'x64' && matrix.go >= '1.17' }}
+        run: |
+          go get github.com/jordanlewis/gcassert/cmd/gcassert
+          gcassert ./...
+
       - name: 'BuildTest for armv7'
         if: ${{ matrix.arch == 'armv7' }}
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -30,18 +34,66 @@ jobs:
...
-        if: ${{ matrix.go != '1.13' }}
+        if: ${{ matrix.arch == 'x64' && matrix.go != '1.13' }}
         run: |
           go get honnef.co/go/tools/cmd/staticcheck
           staticcheck ./...
+
+      - name: 'GCAssert'
+        # Only run gcassert on latest version of Go. Inlining heuristics change
+        # from version to version.
+        if: ${{ matrix.arch == 'x64' && matrix.go == '1.17' }}
```
jordanlewis, line 71: Bleh, I feel like we won't remember to update this.

jordanlewis (approved):

.github/workflows/go.yml
jordanlewis: Maybe change to >= for now? So at least it'll fail if the heuristics change and we get a version update.

yuzefovich (commented):
:lgtm: 

> Reviewed 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/apd/111#-:-Mt4jbH96-Vnx8ROBG0r:bg7igkv)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @jordanlewis, @nvanbenschoten, and @yuzefovich)

.github/workflows/go.yml
nvanbenschoten: Done.

nvanbenschoten:
TFTRs!

PR #116, opened 2022-06-14
apd: remove some allocs from Decimal.setString

josharian:
exps contains exactly 0, 1, or 2 elements.
Provide space for 2 elements up front.
This is an easy, broadly positive optimization.
Some of the time/op benchmarks are noisy;
the allocs benchmarks are clearer.

```
name                           old time/op    new time/op    delta
Exp/P5/S-4/D-2-8                  823ns ± 2%     917ns ±23%  +11.34%  (p=0.000 n=19+19)
Exp/P5/S-4/D2-8                   894ns ±18%     876ns ±15%     ~     (p=0.141 n=18+18)
Exp/P5/S-1/D-2-8                 1.60µs ± 1%    1.73µs ±26%   +8.07%  (p=0.000 n=20+19)
Exp/P5/S-1/D2-8                  1.77µs ± 1%    1.96µs ±21%  +11.08%  (p=0.000 n=20+19)
Exp/P5/S2/D-2-8                  3.92µs ± 1%    4.07µs ±10%   +3.93%  (p=0.049 n=20+17)
Exp/P5/S2/D2-8                   4.38µs ± 0%    4.44µs ± 2%   +1.42%  (p=0.000 n=19+19)
Exp/P10/S-4/D-10-8               1.61µs ± 0%    2.01µs ±57%  +24.95%  (p=0.000 n=17+19)
Exp/P10/S-4/D-2-8                1.42µs ± 0%    1.49µs ± 8%   +4.24%  (p=0.000 n=19+17)
Exp/P10/S-4/D2-8                 1.59µs ± 0%    1.62µs ± 5%     ~     (p=0.587 n=18+20)
Exp/P10/S-4/D10-8                1.54µs ± 0%    1.54µs ± 0%   -0.28%  (p=0.000 n=16+18)
Exp/P10/S-1/D-10-8               2.98µs ± 0%    2.97µs ± 0%   -0.17%  (p=0.002 n=20+19)
Exp/P10/S-1/D-2-8                2.81µs ± 0%    2.81µs ± 0%     ~     (p=0.096 n=20+18)
Exp/P10/S-1/D2-8                 2.99µs ± 0%    3.01µs ± 4%     ~     (p=0.231 n=19+19)
Exp/P10/S-1/D10-8                2.90µs ± 0%    2.96µs ± 7%     ~     (p=0.061 n=20+17)
Exp/P10/S2/D-10-8                6.25µs ± 0%    6.58µs ± 9%   +5.22%  (p=0.000 n=18+20)
[...]

PR #119, opened 2022-06-16
apd: add inline fast-path to BigInt.SetString

nvanbenschoten:
Not important, but easy enough to do and makes a difference for the
performance of `Decimal.setString`, which #116 was interested in.

### DIFF
--- a/bigint.go
+++ b/bigint.go
@@ -886,6 +886,10 @@ func (z *BigInt) SetInt64(x int64) *BigInt {
 
 // SetString calls (big.Int).SetString.
 func (z *BigInt) SetString(s string, base int) (*BigInt, bool) {
+	if i, err := strconv.ParseInt(s, base, 64); err == nil {
+		z.SetInt64(i)
+		return z, true
+	}
 	var tmp1 big.Int //gcassert:noescape
 	zi := z.inner(&tmp1)
 	if _, ok := zi.SetString(s, base); !ok {


josharian:
Nice. Is the speed-up primarily from avoiding big.Int's SetString method? If so, would it be worth fixing upstream instead (or in addition)?

nvanbenschoten:
> Is the speed-up primarily from avoiding big.Int's SetString method? If so, would it be worth fixing upstream instead (or in addition)?

Right. Beyond the reduced complexity of `strconv.ParseInt` compared to `(*BigInt).SetString` thanks to the fixed bit size, the main win here is the avoidance of a heap-allocated `strings.Reader`. I don't see a good way to avoid this in the standard library, given the different callers to `(*BigInt).setFromScanner`. At least not until the standard library adopts generics, and even then I don't recall whether the new GC shape stenciling will be enough to avoid heap allocations in cases like these. Do you happen to know?

josharian:
> I don't see a good way to avoid this in the standard library, given the different callers to (*BigInt).setFromScanner.

I had in mind putting it at the beginning of `(*Int).SetString`, not `(*Int).setFromScanner`, here:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/math/big/int.go;l=425

(Given where we are in the Go release cycle, it'd be worth doing here as well.)

PR #125, opened 2022-10-15
apd: Improve string conversion efficiency

miretskiy:
Prefer to use Append and allocate scratch space
to improve speed and efficiency of string conversion:

```
name              old time/op    new time/op    delta
DecimalString-10     147ns ± 1%      56ns ± 4%  -62.14%  (p=0.000
n=10+10)

name              old alloc/op   new alloc/op   delta
DecimalString-10     84.0B ± 0%     39.0B ± 0%  -53.57%  (p=0.000
n=10+10)

name              old allocs/op  new allocs/op  delta
DecimalString-10      5.00 ± 0%      1.00 ± 0%  -80.00%  (p=0.000
n=10+10)
```

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -24,11 +24,11 @@ import (
 // runBenches benchmarks a given function on random decimals on combinations of
 // three parameters:
 //
-//   precision:    desired output precision
-//   inScale:      the scale of the input decimal: the absolute value will be between
-//                 10^inScale and 10^(inScale+1)
-//   inNumDigits:  number of digits in the input decimal; if negative the number
-//                 will be negative and the number of digits are the absolute value.
+//	precision:    desired output precision
+//	inScale:      the scale of the input decimal: the absolute value will be between
+//	              10^inScale and 10^(inScale+1)
+//	inNumDigits:  number of digits in the input decimal; if negative the number
+//	              will be negative and the number of digits are the absolute value.
 func runBenches(
 	b *testing.B, precision, inScale, inNumDigits []int, fn func(*testing.B, *Context, *Decimal),
 ) {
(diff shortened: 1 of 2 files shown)


bench_test.go
```diff
@@ -103,3 +103,25 @@ func BenchmarkLn(b *testing.B) {
...
+	const numDecimals = 8192
+	corpus := func() []Decimal {
+		res := make([]Decimal, numDecimals)
+		for i := 0; i < numDecimals; i++ {
+			var d Decimal
+			_, err := d.SetFloat64(rng.Float64())
+			if err != nil {
+				b.Fatal(err)
+			}
+			res[i] = d
```
nvanbenschoten, line 118: nit, but can we simplify this to:
```go
for i := range res {
    _, err := res[i].SetFloat64(..)
    if err != nil {
        b.Fatal(err)
    }
}
```

format.go
```diff
@@ -57,7 +57,8 @@ func (d *Decimal) Append(buf []byte, fmt byte) []byte {
 		return append(buf, "unknown"...)
 	}
 
-	digits := d.Coeff.String()
+	var scratch [16]byte
```
nvanbenschoten, line 60: Does `scratch` escape to the heap? If it does, could we Append directly to `buf`? Or temporarily use the spare capacity at the end of `buf`? `fmtE` and `fmtF` look like they might make that tricky.

nvanbenschoten (approved):
Nice improvement!

bench_test.go
miretskiy: Done.

format.go
miretskiy: Doesn't look like it;
I couldn't find scratch itself in the output, but 
```
./format.go:107:35: digits does not escape
```
refers to digits argument to fmtE/f functions -- so, no I don't think scratch escapes.

PR #128, opened 2023-04-25
format: format 0E-x where x > 6 like PostgreSQL does

otan:
Officially in the GDA spec, formatting a 0 with any decimal places should be 0. However, this code doesn't do that - it formats the number of 0s up to 6, and then `0E-exponent` afterwards.

To make this more "PG" compatible, we fix this by ensuring we display as many 0s as the exponent allows in the `.String()` case.

Refs https://github.com/cockroachdb/cockroach/issues/102217, but not 100% sure I like the solution.


rafiss (approved):
lgtm! but nit: since this is a lower level library used by others, i wonder if we should make the commit/comments say "PostgreSQL" instead of PG. ideally we'd not have to mention PG at all, but it sounds like we have to, since PG deviates from the official spec?

otan:
yep, done

PR #131, opened 2024-02-09
apd: add go 1.21 and 1.22 testing to CI

nvanbenschoten:
This commit adds a test variant for go 1.21 and 1.22.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -27,6 +27,8 @@ jobs:
           - '1.18'
           - '1.19'
           - '1.20'
+          - '1.21'
+          - '1.22'
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -71,9 +73,9 @@ jobs:
       - name: 'GCAssert'
         # Only run gcassert on the latest versions of Go. Inlining heuristics
         # change from version to version.
```
petermattis, line 75: This comment seems out of date. We run gcassert on go1.18+.

petermattis (approved):
LGTM

Did you consider removing testing for go1.13 and go1.14? Not sure if it matters. This repo rarely changes.

.github/workflows/go.yml
nvanbenschoten: "the latest versions" was meant to imply that there might be a few recent versions on which we run gcassert, not just the single latest.

nvanbenschoten:
> Did you consider removing testing for go1.13 and go1.14? Not sure if it matters. This repo rarely changes.

Yeah, we should go through and remove old versions from CI, now that they're quite stale. I'll leave that for next time, as they're also not hurting anything.

PR #132, opened 2024-02-12
Allow decompose to make use of the buffer

Gilthoniel:
The comment of Decompose specifies that `buf` can be used to fill the coefficient if the size is big enough however the current implementation simply ignores it. This is problematic when attempting to encode without allocation. This patch attempts to provide the feature.


Gilthoniel:
@petermattis maybe ?

decomposer.go
```diff
@@ -48,9 +48,11 @@ type decomposer interface {
 
 var _ decomposer = &Decimal{}
 
-// Decompose returns the internal decimal state into parts.
-// If the provided buf has sufficient capacity, buf may be returned as the coefficient with
-// the value set and length set as appropriate.
+// Decompose returns the internal decimal state into parts. If the provided buf
```
nvanbenschoten, line 51: We should also update the corresponding comment on `decomposer interface`.

decomposer.go
```diff
@@ -69,7 +71,19 @@ func (d *Decimal) Decompose(buf []byte) (form byte, negative bool, coefficient [
...
+
+	sizeInBytes := (d.Coeff.BitLen() + 8 - 1) / 8 // math.Ceil(d.Coeff.BitLen()/8.0)
+	if cap(buf)-len(buf) >= sizeInBytes {
+		// it extends the buffer as the filling of bytes expects an already
+		// allocated slice.
+		buf = append(buf, make([]byte, sizeInBytes)...)
+
+		// We can fit the coefficient in the given buffer which prevents an
+		// allocation.
+		coefficient = d.Coeff.FillBytes(buf)
```
nvanbenschoten, line 83: `FillBytes` is only available in go1.15 and above. We were meaning to bump the minimum supported version of this library anyway, so that's ok. We'll just need to remove testing for go1.13 and go1.14 in `.github/workflows/go.yml`.

nvanbenschoten: At some point, we can also merge the conditional compilation `bigint_go1.X(_test).go` files away, but that's clean up which we can do separately.

decomposer.go
```diff
@@ -69,7 +71,19 @@ func (d *Decimal) Decompose(buf []byte) (form byte, negative bool, coefficient [
 	// Finite form.
 	negative = d.Negative
 	exponent = d.Exponent
-	coefficient = d.Coeff.Bytes()
+
+	sizeInBytes := (d.Coeff.BitLen() + 8 - 1) / 8 // math.Ceil(d.Coeff.BitLen()/8.0)
+	if cap(buf)-len(buf) >= sizeInBytes {
+		// it extends the buffer as the filling of bytes expects an already
```
nvanbenschoten, line 77: nit: s/it/It/

decomposer.go
```diff
@@ -69,7 +71,19 @@ func (d *Decimal) Decompose(buf []byte) (form byte, negative bool, coefficient [
 	// Finite form.
 	negative = d.Negative
 	exponent = d.Exponent
-	coefficient = d.Coeff.Bytes()
+
+	sizeInBytes := (d.Coeff.BitLen() + 8 - 1) / 8 // math.Ceil(d.Coeff.BitLen()/8.0)
+	if cap(buf)-len(buf) >= sizeInBytes {
```
nvanbenschoten, line 76: > cap(buf)-len(buf)

What does it mean for `len(buf) != 0`? Should we preserve the data in the slice and only use the unused capacity? The contract here, which comes from https://github.com/golang/go/issues/30870, is ambiguous.

The usual convention for when a function like this takes a scratch space buffer is that it is permitted to overwrite any data in the buffer. So the function checks the capacity (`if cap(buf) >= sizeInBytes`), sets the length to the needed length if there's room (`buf = buf[:sizeInBytes]`), and then writes into the buffer from the beginning (`coefficient = d.Coeff.FillBytes(buf)`).

decomposer.go
```diff
@@ -69,7 +71,19 @@ func (d *Decimal) Decompose(buf []byte) (form byte, negative bool, coefficient [
...
+		// it extends the buffer as the filling of bytes expects an already
+		// allocated slice.
+		buf = append(buf, make([]byte, sizeInBytes)...)
```
nvanbenschoten, line 79: Have you verified that the `make([]byte, sizeInBytes)` does not lead to an actual heap allocation. If it does, we'd lose the benefit of this change.

nvanbenschoten (commented):
Hi @Gilthoniel, thanks for the contribution! Out of curiosity, is this an API that you are using or intend to use?

decomposer.go
Gilthoniel: Yeah that makes sense, I will change to this approach instead.

decomposer.go
Gilthoniel: Yes that's highlighted by the benchmark:
```
BenchmarkDecomposerDecompose/no_allocation-8         	73670545	        14.98 ns/op	       0 B/op	       0 allocs/op
BenchmarkDecomposerDecompose/one_allocation-8        	42363051	        27.63 ns/op	       8 B/op	       1 allocs/op
```

Gilthoniel: First one is providing a buffer and the other is providing nil.

decomposer.go
Gilthoniel: Thanks I completely forgot to check for version compatibility. I'm also adding 1.21 and 1.22 if that's ok.

Gilthoniel:
@nvanbenschoten thanks a lot for the review, I think I resolved your points. To answer your question, I'm already using the interface to marshal the decimal for transfers over the network. This is much more efficient than a string encoding and reduce the load on the garbage collector, even more with this PR.

decomposer.go
nvanbenschoten: Is that needed? We did so in https://github.com/cockroachdb/apd/pull/131.

decomposer.go
nvanbenschoten: Since you made that change, I don't understand why the `buf = append(buf, make([]byte, sizeInBytes-len(buf))...)` logic is needed, instead of just `buf = buf[:sizeInBytes]`.

decomposer.go
nvanbenschoten: Is that needed? We did so in https://github.com/cockroachdb/apd/pull/131.

Gilthoniel: Very true, I'll remove the change here.

decomposer.go
Gilthoniel: Yes you're right that's not necessary. I actually learned today that it's allowed to increase the length if the capacity is big enough without append.

Gilthoniel:
@nvanbenschoten ping

nvanbenschoten (approved):
Looks great, thanks @Gilthoniel.

PR #143, opened 2025-12-04
fix negative zero from multiplication/division by zero bigint

mhaddon:
(-1) * 0 was producing a "negative zero" where Sign() returns -1 but String() returns "0". This breaks Cmp() comparisons.

addInline already handles this correctly - just needed the same fix in mulInline, quoInline, and remInline.

```go
result := new(apd.BigInt).Mul(apd.NewBigInt(-1), apd.NewBigInt(0))

result.String() // "0" 
result.Sign()   // -1 (should be 0)
result.Cmp(apd.NewBigInt(0)) // -1 (should be 0)
```

Same issue with division and remainder when the result is zero.

Added the same zero check to all three functions. Existing tests pass.

### DIFF
--- a/bigint.go
+++ b/bigint.go
@@ -325,7 +325,7 @@ func addInline(xVal, yVal uint64, xNeg, yNeg bool) (zVal uint64, zNeg, ok bool) 
 //gcassert:inline
 func mulInline(xVal, yVal uint64, xNeg, yNeg bool) (zVal uint64, zNeg, ok bool) {
 	hi, lo := bits.Mul64(xVal, yVal)
-	neg := xNeg != yNeg
+	neg := xNeg != yNeg && lo != 0
 	overflow := hi != 0
 	return lo, neg, !overflow
 }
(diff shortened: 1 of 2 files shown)


bigint.go
```diff
@@ -326,6 +326,9 @@ func addInline(xVal, yVal uint64, xNeg, yNeg bool) (zVal uint64, zNeg, ok bool)
 func mulInline(xVal, yVal uint64, xNeg, yNeg bool) (zVal uint64, zNeg, ok bool) {
 	hi, lo := bits.Mul64(xVal, yVal)
 	neg := xNeg != yNeg
```
michae2: nit: This could be shortened to `neg := xNeg != yNeg && lo != 0` (and same with a couple others).

michae2 (commented):
Thanks for the fix! Would you be willing to add a short test to bigint_test.go?

mhaddon:
Is that better?

michae2 (approved):
Looks great! Thank you!

Assuming tests pass, I'll go ahead and merge.

Issue #144, opened 2026-03-12
setString accepts sign characters after decimal point, producing corrupt Decimal values


jms-bc:
It has previously been raised in #120 that `SetString` will panic on `".-400000000000000000000000000000000000000"`. During fuzzing we have discovered that similar inputs via `NewFromString` can produce corrupt decimal values that result in invalid arithmetic operations.

This is triggered by a value like `".-1"` that has (1) no digits before the decimal point, and (2) a sign character after the decimal point. This is due to `Decimal.setString`'s reliance on the downstream parser in `BigInt.SetString` to handle validation of the mantissa, but they have different sets of valid values.

## Cause

`setString` removes the decimal point by concatenating the characters before and after it. If a sign character (`-` or `+`) appears immediately after the dot, it becomes the leading character of the string passed to `BigInt.SetString` which accepts it. The sign check in `setString` only inspects the start of the string, so a sign after the dot is never rejected.

If the sign is negative, this produces a `Decimal` with a negative `Coeff`, which should never happen and results in a corrupted `Decimal` with `Sign()` and `Cmp()` reporting the decimal's positive sign from `Decimal.Negative` whilst arithmetic operations use the negative coefficient.

If the sign is positive, `BigInt.SetString` absorbs the `+` silently and produces the correct coefficient. However the `+` character is counted as a fractional digit when computing the exponent which is correspondingly off-by-one. This produces a value 10x smaller than intended, i.e. `.+5` parses as `0.05` rather than `0.5`.

## Proposed Fix

The mantissa should be validated before being passed to `BigInt.SetString` - after the removal of the decimal point and exponent, the remaining string should contain only ASCII digits:

```
// Accept only ASCII characters in the range 0-9
for _, c := range s {
    if c < '0' || c > '9' {
        return 0, fmt.Errorf("could not parse mantissa: %s", s)
    }
}
```

We will file a PR shortly with this fix.

## Reproduction

Some sample code to reproduce this behaviour is:

```
package main

import (
        "fmt"
        "log"

        apd "github.com/cockroachdb/apd/v3"
)

func main() {
        negativeCoeff()
        offByOneExponent()
}

// negativeCoeff demonstrates that ".-1" produces a Decimal whose Coeff is
// negative, causing Sign() and Cmp() to disagree with arithmetic operations
// that use the raw coefficient.
func negativeCoeff() {
        fmt.Println("=== Case 1: negative coefficient (input \".-1\") ===")

[...]

Issue #120, opened 2022-06-17
panic with SetString(".-400000000000000000000000000000000000000")


josharian:
This test panics:

```go
func TestSetStringMalformed(t *testing.T) {
	c := &BaseContext
	d := new(Decimal)
	tests := []string{
		".-4",
		".-400000000000000000000000000000000000000",
	}
	for _, test := range tests {
		_, _, err := c.SetString(d, test)
		if err == nil {
			t.Errorf("expected error for SetString(%q)", test)
		}
	}
}
```

The panic is in the second test case. Also, I think `".-4"` should yield an error (the first test case).

---

This was found (indirectly) by go-fuzz. Would there be interest in adding a fuzz function or three directly? (I don't have much time to dedicate to this, but I can at least give it a nudge, and/or continue my indirect fuzzing.)

nvanbenschoten:
> Would there be interest in adding a fuzz function or three directly?

Yes, any contribution (large or small) to improve the testing of this library would be appreciated.

odeke-em:
I also just re-discovered this panic while fuzzing [gnolang/gnovm](https://github.com/gnolang/gno) with apd.Decimal.UnmarshalText

```go
v := new(apd.Decimal)
in := []byte(".-700000000000000000000000000000000000000")
_ = v.UnmarshalText(in)
```

 Kind cc @jaekwon

yuzefovich:
this has been fixed via #145

PR #145, opened 2026-03-12
apd: prevent NewFromString parsing invalid decimals

jms-bc:
Fixes a parsing bug outlined in #144  where sign characters (`+`, `-`) immediately after a decimal point bypass `setString`'s sign validation and are passed to `big.Int.SetString`, which accepts them. This produces corrupt `Decimal` values.

The fix validates that the mantissa contains only ASCII digits (0-9) after the decimal point and exponent have been removed, before it is passed to `big.Int.SetString`.

Related: #120


mw5h (approved):
:lgtm:

Thanks for this fix!

